PR #7, opened 2021-07-12
update config for using the param store values

charleystran:
This PR updates the config files so that production can pull in DB credentials from env vars that come from param store. This PR is related to https://github.com/department-of-veterans-affairs/lighthouse-platform-backend-deployment/pull/1


config/database.yml
```diff
@@ -1,27 +1,11 @@
...
-#
 default: &default
   adapter: postgresql
   encoding: utf8
   pool: 5
   timeout: 5000
-  host: <%= ENV['DB_HOST'] %>
-  username: <%= ENV["POSTGRES_USER"] %>
-  password: <%= ENV["POSTGRES_PASSWORD"] %>
+  host: <%= ENV['DATABASE_HOST'] %>
```
wwanamaker, line 6: Is there a reason this one is in single quotes?

FonzMP: I think rails convention is single quote unless you are using interpolation. not sure about the .yml files within rails though. Probably a nice mix of the two.

Wonder if there's something rubocop could do with this to help us stay consistent?

FonzMP:
I'm good with the changes once we get the build to succeed. Think the change back to docker-compose will fix the current issue.

config/database.yml
charleystran: Single quotes are right, I just didnt update all the old values

gileswells: @charleystran if single quotes are proper then you've got new lines right below this that should be updated as well.

docker-compose.yml
```diff
@@ -12,19 +12,19 @@ services:
...
   db:
     image: postgres:13.1
     expose:
       - "5432"
     ports:
       - "5432"
     environment:
-      - POSTGRES_USER=postgres
-      - POSTGRES_PASSWORD=postgres
-      - DB_HOST=db
```
FonzMP, line 27: PG has some required environment variables (well, only 1 required) and specific naming for the values we want to set. I think we'll need to revert these for PG to work correctly.
https://hub.docker.com/_/postgres

PR #10, opened 2021-07-15
add more healthcheck logging

charleystran:
add more logging for the healthcheck endpoint


FonzMP (approved):
:shipit:

config/initializers/health_check.rb
```diff
@@ -12,7 +12,7 @@
 
   # Disable the error message to prevent /health_check from leaking
   # sensitive information
```
dsteeve, line 14: this change goes against the documented comment.  what type of sensitive info could be leaked?

charleystran: hey @dsteeve  Thank you for noticing. Currently nothing would get exposed. The change has been made to increase debugging for our initial deploy to the DVP infra. There have been some issues and we are just looking for more verbose logging. Currently the app is almost vanilla as we work on the deploy so there is no risk of any exposure.

PR #11, opened 2021-07-24
Gh alert refactor and prod config

charleystran:
This PR's goal is to bring the repo back to a real prod build config. As as result, some changes were needed to the GH alert code to get the specs happy


FonzMP:
getting a few failing tests locally. not sure exactly why as it should be hitting the POST, but giving me a `Method Not Allowed / 405` on the tests.... I think we want to switch the `build.sh` to include `ci`, correct? Running with that change locally exposes the failing tests. The build doesn't seem to be doing so now.
`docker-compose run app bundle exec rails db:create lint`

app/services/github_alert_creator.rb
```diff
@@ -4,15 +4,26 @@
...
 
   def initialize(body)
     @body = body
+    @secret_type = body.dig(:alert, :secret_type)
+    @full_name = body.dig(:repository, :full_name)
+    @html_url = body.dig(:repository, :html_url)
+    @login = body.dig(:organization, :login)
   end
 
   # rubocop:disable Metrics/MethodLength
```
FonzMP, line 24: do we want to move all rubocop items in to the .rubocop.yml? This allows us the ability to review from a centralized point as opposed to spreading it throughout the entire application.

charleystran: this is a specific exception so it should stay in the file

FonzMP:
LGTM 🚀

PR #12, opened 2021-07-28
Api 7674 apply workflow

FonzMP:
This PR introduces the beginning phase of setting up the lighthouse-platform-backend to replace dynamoDB for consumer information. The goal for this PR is to try and mimic the current developer portal apply flow as best as possible. This application will run in parallel to the current developer-portal-backend process as we begin next steps of shifting our dynamoDB entries into the new lighthouse-platform-backend. With this process, we're also implementing a method of updating an existing entry, reducing the duplication of consumer data, which will in turn provide a much more easily managed and understandable collection of data.


app/models/consumer.rb
```diff
@@ -0,0 +1,35 @@
...
+  accepts_nested_attributes_for :consumer_api_assignments
+
+  before_save :manage_tos
+  before_save :manage_apis
+
+  private
+
+  def manage_tos
+    if new_record? && tos_accepted == 'false'
+      raise 'TOS not accepted'
```
charleystran, line 21: Lets not raise an error here, versus, lets do error.add(:tos_accepted, 'some message') that way in the response from the controller it is properly returned in the wrrors.full_messages

FonzMP: changed this up a bit. curious on your take.

config/routes.rb
```diff
@@ -5,6 +5,7 @@
   # everything must scoped under platform-backend
   scope '/platform-backend' do
     get 'admin/dashboard', to: 'admin/dashboard#index'
+    resources :consumers
```
charleystran, line 8: Lets limit this to only create for now. so something like resources :consumers, only: [:create] since we are not yet using the other methods but we will eventually

charleystran (approved):
As long as the build passes, let er rip! :shipit:

PR #13, opened 2021-08-03
remove open_api_url validation

FonzMP:
Remove presence validation for open_api_url as it may not exist.


FonzMP:
This is accomplished here:
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/15

PR #15, opened 2021-08-09
Load initial apis

charleystran:
This PR is set to take in a file and a base host to import our current list of APIs that exist on the gateway. The CSV that will be parsed will be shared on keybase and is an argument in the rake file


spec/requests/admin/api/v0/apis_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+          auth_method: 'key_auth',
+          environment: 'sandbox',
+          open_api_url: 'https://sandbox-api.va.gov/services/claims/docs/v0/api',
+          base_path: '/services/claims/v0',
+          service_ref: 'somerandom-guid',
+          api_ref: 'claims'
+        }
+      }
+    end
+    # rubocop:enable Metrics/MethodLength
```
FonzMP, line 21: we have a `valid_params` defined in `consumers_request_spec.rb` as well without disabling rubocop. I think we should define how we want to build this out now before we get much further in. Here's three thoughts:
1. `consumers_requqest_spec.rb` has `let :valid_params do` and that doesn't require `rubocop:disable`, so we could modify this to match
2. add `valid_params` to `rubocop.yml` -> `Metrics/MethodLength` -> `IgnoredMethods` and use this format in all tests (this helps keep specs clean)
3. keep this as is and change `consumers_request_spec.rb` to match.

For reference: https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/f56aa3177d84cd454620dbfc4b0fcb39acb3a3ba/spec/requests/consumers_request_spec.rb#L7

rakelib/load_initial_apis.rake
```diff
@@ -0,0 +1,26 @@
...
+  host = args.extras.last
+  url = "#{host}/platform-backend/admin/api/v0/apis"
+  apis_list = CSV.parse(File.read(file_path), headers: true)
+  puts apis_list.inspect
+  apis_list.each do |api|
+    Rails.logger.warn "------- #{api['api_name']} -------"
+    params = {
+      api: {
+        name:	api.first.last,
+        version: api['version'].to_i,
```
FonzMP, line 16: do we need `to_i` since this field is a string?

rakelib/load_initial_apis.rake
```diff
@@ -0,0 +1,26 @@
...
+        auth_method: api['auth_method'],
+        environment: api['environment'],
+        open_api_url: api['open_api_url'],
+        base_path: api['base_path'],
+        service_ref: api['service_ref']
+      }
```
FonzMP, line 22: missing `api_ref` which also causes issues as of current cause we're scoping it to `api_ref` and `environment`, but we share those with `version`. For reference, the 2nd API on our sheet. `claims` -> `sandbox` version `0` and `1` causes a conflict.

app/controllers/admin/api/v0/apis_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    if api.save
+      render json: api, serializer: ApiSerializer
+    else
+      render json: { error: api.errors.full_messages }, status: :unprocessable_entity
+    end
+  end
+
+  # rubocop:disableMetrics/MethodLength
+  def bulk_upload
+    # rubocop:disableRails/SkipsModelValidations
```
FonzMP, line 17: aren't the rubocop lines here redundant? we have them in the `.rubocop.yml`.

.rubocop.yml
```diff
@@ -80,16 +80,21 @@ Layout/LineLength:
 Metrics/MethodLength:
   Exclude:
     - 'rakelib/support/shell_command.rb'
+    - 'app/controllers/admin/api/v0/apis_controller.rb'
   IgnoredMethods:
     - 'user_params'
-
+    - 'bulk_upload'
```
FonzMP, line 86: I believe we could either keep 83 or 86. do we need both? I believe 83 skips the file, and 86 skips the method. Think it would be best to keep 86 and allow rubocop to watch the remainder of the file.

app/controllers/admin/api/v0/apis_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+        base_path: api.dig('api', 'base_path'),
+        service_ref: api.dig('api', 'service_ref'),
+        created_at: Time.zone.now,
+        updated_at: Time.zone.now
+      }
+    end, unique_by: [:service_ref])
+    render json: { data: 'apis_uploaded' }
+    # rubocop:enableRails/SkipsModelValidations
+  end
+  # rubocop:enableMetrics/MethodLength
```
FonzMP, line 34: same as above, could remove rubocop lines.

charleystran: removed

app/controllers/admin/api/v0/apis_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  def create
+    api = Api.new api_params
+    if api.save
+      render json: api, serializer: ApiSerializer
+    else
+      render json: { error: api.errors.full_messages }, status: :unprocessable_entity
+    end
+  end
+
+  # rubocop:disableMetrics/MethodLength
```
FonzMP, line 15: this was left in. the enableMetrics was removed though.

FonzMP:
I'm good with this functionally, just think we need to clean up rubocop items. We're doing `.rubocop.yml` changes for some and rubocop "comments" for others. We're also duplicating some rules. Think we should fine tune our approach before it gets convoluted.

FonzMP:
LGTM 🚀

PR #16, opened 2021-08-12
add removal and unique build name to docker compose

charleystran:
This PR is an attempt to fix the issue of containers inside of Jenkins from using cached images. This will tag the images we create with the build number and make sure that the docker compose down removes the image.


FonzMP:
:shipit: LGTM

PR #17, opened 2021-08-27
[API-7681] Initial injection of the portal UI

gileswells:
https://vajira.max.gov/browse/API-7681

Initial add of the admin UI which mirrors the current developer portal.


FonzMP:
I'm good with these once we get the build working.

PR #19, opened 2021-09-01
update gems for brakeman

FonzMP:
updating gems to pass brakeman scan. this PR in particular focuses on the actionpack vulnerability from 6.1.3.2.


charleystran (approved):
LGTM

PR #20, opened 2021-09-02
Api 7673 dynamo data imported

charleystran:
This PR implements the import and de-duplication of the dynamo database into the postgres database.


Gemfile
```diff
@@ -5,12 +5,14 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
...
 ruby '3.0.0'
 
+gem 'aws-sdk-dynamodb', '~> 1.62'
 gem 'bootsnap', '>= 1.4.4', require: false
 gem 'devise'
 gem 'figaro'
 # gem 'health_check'
 gem 'jbuilder', '~> 2.7'
 gem 'jsonapi-serializer'
+gem 'oktakit', git: 'https://github.com/charleystran/oktakit', branch: 'add_authorization_servers'
```
SlakrHakr, line 15: is there any plan to get this merged back upstream into `Shopify/oktakit` ?

i see the original commit to this fork to "add authorization servers" was made back in October 2020:
https://github.com/charleystran/oktakit/commit/5553d815e15602b375d26cc025f81263d18dced8
which doesnt seem to have anything we wouldnt want to contribute back but 🤷🏻

app/models/consumer.rb
```diff
@@ -15,6 +15,25 @@ class Consumer < ApplicationRecord
   before_save :set_tos
   before_save :manage_apis
 
+  def self.create_from_import(consumer, params)
+    consumer.description = params['description']
+    consumer.organization = params['organization']
+    consumer.sandbox_gateway_ref = params['sandbox_gateway_ref'] if params['sandbox_gateway_ref'].present?
+    consumer.sandbox_oauth_ref = params['sandbox_oauth_ref'] if params['sandbox_oauth_ref'].present?
+    consumer.tos_accepted_at = Time.zone.now
```
SlakrHakr, line 23: I'm assuming this is a new thing that the current workflow (dynamodb, etc. doesnt currently support?)
I assume consumers do "accept" a terms of service but we just dont store the time they do?
Which, assuming all that above is correct, why do we care to store the time the consumer accepted the terms of service now?

charleystran: correct, they just have an accepted field that we store. We are storing the time now  because we can treat time a boolean, and might need to inform consumers based on given date they accepted. An idea in the past was having consumers re-review the TOS and we could track the date.

app/models/consumer.rb
```diff
@@ -15,6 +15,25 @@ class Consumer < ApplicationRecord
   before_save :set_tos
   before_save :manage_apis
 
+  def self.create_from_import(consumer, params)
```
SlakrHakr, line 18: I might very well be missing something but I don't see this referenced anywhere outside of specs. Was this intended to be used in this commit or is this here for some future use?

app/controllers/consumers_controller.rb
```diff
@@ -32,6 +32,7 @@ def user_params
         prod_gateway_ref
         prod_oauth_ref
         apis_list
+        apis
```
SlakrHakr, line 35: why the addition of `apis` here when `apis_list` is already permitted? are they different? why would a consumer send in one vs the other?

app/models/consumer.rb
```diff
@@ -15,6 +15,25 @@ class Consumer < ApplicationRecord
...
+    consumer.sandbox_oauth_ref = params['sandbox_oauth_ref'] if params['sandbox_oauth_ref'].present?
+    consumer.tos_accepted_at = Time.zone.now
+    consumer.tos_version = Figaro.env.current_tos_version
+    consumer.save
+
+    manage_apis(consumer, params[:apis_list])
+  end
+
+  def self.manage_apis(consumer, apis_list)
+    apis_list.split(',').each do |api_name|
```
SlakrHakr, line 31: does this need to check for nil like the `manage_apis` method on line 47 does?

sidenote: I know it's not in this particular pr but the line 47 `manage_apis` returns an array it looks like, so shouldn't it return an empty array on line 48 if `apis_list` is blank?

why is the logic between this method and the line 47 `manage_apis` duplicated but slightly different? can they be merged and have one call the other?

```
def self.manage_apis(consumer, apis_list)
    return [] if apis_list.blank?
    
    apis_list.split(',').map do |api|
      api_model = Api.find_by(api_ref: api.strip, environment: 'sandbox')
      apis << api_model if api_model.present? && api_ids.exclude?(api_model.id)
    end
end

def manage_apis
   Consumer.manage_apis(self, apis_list)
end
```

Gemfile
SlakrHakr: conversation offline that I'll just document here:

@charleystran says the plan is to get these changes back into `Shopify/oktakit` but there is some testing and mocking stuff required before that can happen.

I'd argue it'd be valuable to do sooner rather than later personally. Never love using branches of things long-term in production-level projects.

spec/services/consumer_import_service_spec.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe KongService do
+  describe 'importing consumers from dynamo and Kong' do
+    it 'imports the the users' do
```
SlakrHakr, line 7: the*

spec/services/dynamo_service_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+            'okta_application_id' => { 'S' => 'yJQfQjbNECoiFVVEXXRG' },
+            'organization' => { 'S' => 'School' },
+            'oAuthRedirectURI' => { 'S' => 'https://fake-redirect.com' },
+            'description' => { 'S' => 'no description' }
+          }
+        ]
+      }
+    end
+
+    it 'imports the the users' do
```
SlakrHakr, line 41: the*

app/controllers/consumers_controller.rb
charleystran: apis should be removed as it became a relational model. Good catch!

spec/services/consumer_import_service_spec.rb
charleystran: fixed

spec/services/dynamo_service_spec.rb
charleystran: fixed

app/models/consumer.rb
charleystran: This was removed as the logic is in the UserService now

app/models/consumer.rb
charleystran: this was removed. It was a potential refactor but ended up not making sense. The code wasnt removed before PR

Gemfile
charleystran: The issue with the tests is having an Okta instance we can VCR that doesnt have real user data. Goes back the issue we have of having a real dev instance with not real user data or anonymizing it

SlakrHakr: why cant we just generate recordings and then manually remove any potential PII? that recording isnt going to change all that much is it?

SlakrHakr (approved):
I still think we should do everything within our power to get our changes back into oktakit proper, but i won't hold up the review just for that work 👍🏻 

🚀

PR #25, opened 2021-10-04
Api 9929 investigate ssl issue

FonzMP:
This PR introduces the health-apis centos8 image which provides access to needed certs. This seems to be correcting the SSL issue as well as a change to only target the production stages of the build.

Notable changes:
1. `assets:precompile` includes a `yarn install` so that's covered. this caused us to run `yarn install` multiple times. The `--silent` is to try and quiet down some chatter in the build as there is A LOT of action currently. This is the same as `./configure --silent` on the ruby install. We could probably minimize chatter further, but still searching.
2. `--target prod` allows us to only build `base` and `prod` in the final `docker build`. We were building all 3 stages.
3. looking in to `shared-mime-info` in `yum install` to cut down on size (potentially)
4. Update in `Gemfile.lock` is to pass brakeman in `ci`.
5. We don't use `retry` currently so is it worth keeping as a proactive measure?


docker-bin/retry
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+for i in {0..5}; do
+  $@ && exit 0 || sleep ${RETRY_SLEEP:-5}
+done
+
+exit 1
```
SlakrHakr, line 7: > We don't use retry currently so is it worth keeping as a proactive measure?

id vote we either use it or remove this file. my first thought, before re-reading this pr's description was "this must be used somewhere... what am i missing..."
id vote just removing the file at this point i guess, assuming we're seeing consistent successes now with these other changes

SlakrHakr (approved):

PR #30, opened 2021-10-22
API-9949: Run the LPB Consumer Migration Process

SlakrHakr:
Finalize process to migrate existing consumer structure over to LPB.
https://vajira.max.gov/browse/API-9949

image


imcgonigle (approved):

app/services/kong_service.rb
```diff
@@ -6,10 +6,16 @@
 class KongService
   attr_reader :client
 
+  #
+  # Note: Use SOCKS Proxy client for any connections except local, which uses standard Net::HTTP
```
FonzMP, line 10: do we want to remove this comment? it was brought over from platform tools `kong_consumers` tooling and I know it caused some confusion when you saw it. Is it worth removing?

spec/rails_helper.rb
```diff
@@ -29,6 +29,7 @@ def self.all_matches
   c.cassette_library_dir = 'spec/support/vcr_cassettes'
   c.hook_into :webmock
   c.default_cassette_options = { decode_compressed_response: true }
+  c.allow_http_connections_when_no_cassette = true
```
FonzMP, line 32: do we want this to be as open as this provides? We can limit via `ignore_localhost` and also set stipulations based on ports, etc;  Just curious if we're letting this be to open. I feel like this is essentially removing VCR with this setting, unless we have a recorded cassette (if I'm understanding correctly). Or are we fine with this since we're working towards testing against local instances?

SlakrHakr: i went with `allow_http_connections_when_no_cassette` here because webmock should already be preventing non-localhost requests:
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/0e40b21f917a3935d7c7a50f4e0f9de7665d6272/spec/rails_helper.rb#L13

that being said, i dont 100% remember the issue i was running into that made me add this line to VCR, so i can definitely look into this a bit and play around with some of the configurations

app/services/kong_service.rb
SlakrHakr: https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/30/commits/ce43ecd8da65a400ec52552b2831d5e4e3316077

spec/rails_helper.rb
SlakrHakr: ok, this is using the same logic from the kong changes commit now, so no longer fully open
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/30/commits/a9980d813319fe91bc5f84e5e9ead8aad8619017

SlakrHakr:
@FonzMP: build is passing again and i believe ive responded to everything 👍🏻

PR #31, opened 2021-10-27
API-9406  Add ability to sign in using GitHub

imcgonigle:
This PR adds the ability for users to sign in using GitHub


SlakrHakr (approved):

app/controllers/admin/dashboard_controller.rb
```diff
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 
 class Admin::DashboardController < ApplicationController
+  before_action :authenticate_user!
+
   def index
     @apis = Api.kept
-    @users = User.kept
+    @users = User.select { |user| user.consumer.present? }
```
SlakrHakr, line 8: it might not be a huge deal at this stage, but id be curious to see how `select` operates if the user has been discarded instead of deleted. thats what `kept` was for there, to filter out any discarded user records. 
```
User.kept.select { |user| user.consumer.present? }
```
would probably work the same as you have it there while also filtering out previously discarded users.

imcgonigle: That makes sense. I just updated it

README.md
```diff
@@ -1,18 +1,56 @@
...
+1. Copy `config/application.yml.sample` to `config/application.yml` and set the env variables from GitHub and Okta
+1. Run `docker-compose up -d` to bring up dependencies
+1. Install gems with `bin/bundle install`
+1. Create the database by running `bin/rake db:create`
+1. Run the migrations with `bin/rake db:migrate`
+1. Seed the database with `bin/rake db:seed`
+1. Start the app by running `bin/rails server`
+
+### Setting Up GitHub Auth
+Lighthouse Platfrom Backend uses GitHub authentication so you'll need to setup an OAuth app, an organization and a team in GitHub before everything will work properly.
```
SlakrHakr, line 17: Platform*

SlakrHakr:
LGTM 🚀

PR #32, opened 2021-11-03
Apply local kong to ci

FonzMP:
This PR sets the current Kong tests to be run against the Kong image within Docker. This will establish an accurate representation of the live environments versus testing against a prerecorded cassette using VCR. I've added a few generalized routes and service methods to incorporate within the tests for validation.


config/environments/test.rb
```diff
@@ -66,6 +66,5 @@
   ENV['DYNAMO_TABLE_NAME'] = 'mocktable'
   ENV['OKTA_TOKEN'] = 'mocktoken'
   ENV['OKTA_API_ENDPOINT'] = 'https://deptva-eval.okta.com/api/v1'
-  ENV['SOCKS_HOST'] = 'http://localhost'
-  ENV['KONG_ELB'] = 'http://api-gateway-dev.vfs.va.gov'
+  ENV['kong_elb'] = 'http://kong:8001'
```
imcgonigle, line 69: `kong_elb` could be switched to  `KONG_ELB` for consistency

imcgonigle (commented):
This looks good. I'm going to run it locally then I'll approve

config/environments/test.rb
FonzMP: I'm curious to see this in action in your local. For some reason I get a failure and it falls back to localhost:4001 in local if I do not modify this. It's like it's case sensitive. Lee said he didn't experience the same issue, so I'm glad to convert this and let the CI pass and change it for consistency, but I will have to find a way to fix my local I guess?

FonzMP: i stand corrected.

imcgonigle (approved):
Looks good. It works on my machine

PR #36, opened 2021-11-12
remove docker port binding

SlakrHakr:


FonzMP (approved):
Good if we need this. :shipit:

PR #39, opened 2021-12-02
Remove old secret from env samples

imcgonigle:
This PR remove a secret that shouldn't have been committed... The old secret has been deleted

### DIFF
--- a/config/application.yml.sample
+++ b/config/application.yml.sample
@@ -8,8 +8,9 @@ kong_password:
 okta_token:
 okta_api_endpoint:
 
-# Github
+# Github OAuth App Secrets
 github_app_id: e2583634c192807ccc28
-github_app_secret: a5b762d35c340f617e54b9c778514c9a9563dd79
+github_app_secret:
+
 # Only used in development
 github_team_id: "5268420"


FonzMP (approved):
LGTM :shipit:

gileswells (dismissed):
This should be fixed with a forced push that rewrites the git history. It'll mess with any active PRs so if we can wait till after Alex finishes his thing that would likely be a very good thing for him. In the end this is a publc repo so this is out there already and archived most likely.

https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository

FonzMP:
> This should be fixed with a forced push that rewrites the git history. It'll mess with any active PRs so if we can wait till after Alex finishes his thing that would likely be a very good thing for him. In the end this is a publc repo so this is out there already and archived most likely.
> 
> https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository

@gileswells FYSA - I believe he went in and changed the app secret prior to this PR, so I'm not sure we'll need the above. If we're concerned with removing the previous value I could see this, but I think we're good with just removing the old secret from the sample cause it's no longer in a functional state.

gileswells:
I feel like the history rewrite is the better solution despite it being inactive. That being said, with it being inactive this method is acceptable.

PR #40, opened 2021-12-06
Api 11432 handle kong interactions

FonzMP:
The initial goal of this PR was to introduce the Kong flow of the Consumer Signup coming from the Developer Portal. It should replicate the request and response needed to match the current signup in respect to Kong. 

To mock a consumer apply, you can hit `/platform-backend/apply` with the below:
```
{
    "apis":"fancy",
    "description":"Development Test",
    "email":"fancy_tester@our.development.com",
    "firstName":"Fancie",
    "lastName":"Pants",
    "organization":"Slacks Express",
    "termsOfService":true
}
```

Please make sure to upload the apis_list.csv found within `/spec/support/` - you can do so using the `/platform-backend/admin/dashboard`.

This PR also reworks the API schema and services associated with the aforementioned. This probably should have been split into it's own PR but I like a challenge.


config/application.yml.sample
```diff
@@ -1,4 +1,4 @@
-dynamo_endpoint: "http://localhost:5000"
+dynamo_endpoint: "http://localhost:4500"
```
FonzMP, line 1: with the recent apple update we have a conflict running on 5000. Just moved to 4500 randomly.

app/models/environment.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Environment < ApplicationRecord
+  has_many :api_environments, dependent: :destroy
+
+  validates :name, presence: true
+end
```
SlakrHakr, line 7: any new model should include Discard::Model so the records can be discarded instead of destroyed. i went ahead and added them in my branch off of this and im close but i have this strange issue where each time i "delete all apis" and then reupload them i get twice as many as i had before. i think its duplicating the record that bridges an api to an environment? still looking into it though

SlakrHakr: ok, these changes seem to fix the admin dashboard functionality as far as i can tell
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/commit/d3e893df81ab3cbc656e0d8200ad2c6609109219

app/models/consumer.rb
```diff
@@ -39,7 +39,8 @@ def manage_apis
     return if apis_list.blank?
 
     apis_list.split(',').map do |api|
-      api_model = Api.find_by(api_ref: api.strip, environment: 'sandbox')
+      api_id = ApiRef.find_by(name: api.strip)['api_id']
+      api_model = Api.find(api_id)
```
SlakrHakr, line 43: is the desire for this to raise an exception or to just move on to the next api in the list if something is wrong?
line 42 and 43 will raise exceptions if the data isnt present in the right way, but then line 44 checks if api_model is present?
i'm assuming line 42 and 43 should protect from exceptions so that apis later in the list still get operated on even if an earlier api in the list causes a problem?

FonzMP: I think I'm following you here. Once we upload the APIs we SHOULD never see this, but there is a chance we get an API added to the developer-portal that doesn't get added to the APIs here (or something as simple as a typo). I would imagine we'd want to just skip any invalid API and continue through the remaining list to prevent any errors, but I'm totally open for discussion on that. 

Side notes:
If we do follow that process, I would imagine that we would send a response with the list of APIs, but if a user signing up does not pay attention, they would not notice that an API had not been included and would be contacting support with questions on why they do not have access. 
It's probably best to prevent the entire signup flow from running, but that being said, we would have to do some quick work to make the changes to fix the signup page, the DB here, or whatever is causing the issue.

app/services/api_service.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class ApiService
+  def gather_apis(api_list)
+    applied_for_apis = api_list.split(',')
+    Api.all.filter do |api|
```
SlakrHakr, line 6: Probably want `Api.kept` here instead of `all`, assuming this wants to ignore previously discarded records.

app/models/consumer.rb
SlakrHakr: sounds like it might be better to leave this as is for now. sounds good to me 👍🏻

app/services/api_service.rb
FonzMP: modified.

SlakrHakr (approved):
LGTM 🚀

PR #45, opened 2022-01-10
Disable GitHub Auth

FonzMP:
Creates a toggle for Github Auth

### DIFF
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -11,8 +11,6 @@ class Admin::DashboardController < ApplicationController
   private
 
   def handle_authenticate
-    # needed to pass rspec
-    # rspec does not seem to recognize ENV changes in before_action
     return if Figaro.env.enable_github_auth.blank?
 
     authenticate_user!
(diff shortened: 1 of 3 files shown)


app/controllers/admin/dashboard_controller.rb
```diff
@@ -1,10 +1,20 @@
...
   def index
     @apis = Api.left_joins(:api_ref, :api_environments).kept
     @users = User.kept.select { |user| user.consumer.present? }
   end
+
+  private
+
+  def handle_authenticate
+    # needed to pass rspec
+    # rspec does not seem to recognize ENV changes in before_action
```
SlakrHakr, line 15: im not convinced this is exclusively rspec. i think rspec is just displaying it. id bet it has something to do with rails underlying implementation of before_action or something 🤷🏻

FonzMP: do you think it's worth removing then, just so we don't cause confusion. I may leave line 14 just in case future folk are questioning our before_action, but perhaps 15 is unnecessary?

SlakrHakr: eh, if it were solely up to me id probably remove both lines, the tests are there and will fail if anyone tries to restructure this

that being said, its not the end of the world there, mostly just wanted to point out that i dont think its exclusively rspec that is the issue here

SlakrHakr (approved):

PR #48, opened 2022-01-20
API-12392 remove npm packages and tighten everything up a bit

SlakrHakr:
https://vajira.max.gov/browse/API-12392

Removes unnecessary npm packages
Removes unnecessary api endpoints
Introduce grape for api endpoints


Dockerfile
```diff
@@ -24,31 +23,30 @@ RUN curl -sL "https://cache.ruby-lang.org/pub/ruby/${RUBY_MAJOR_VERSION}/ruby-${
...
-  -in /etc/pki/ca-trust/source/anchors/VA-Internal-S2-RCA1-v1.cer \
-  -out /home/ruby/va-internal.pem
-ENV NODE_EXTRA_CA_CERTS=/home/ruby/va-internal.pem
+# ----
+
+FROM base AS prod
+
+RUN bundle install --jobs 5 --without development test
+COPY . .
+RUN RAILS_ENV=production SECRET_KEY_BASE=DEFAULT_VALUE_OVERRIDE_AT_RUNTIME bundle exec rails assets:precompile
```
FonzMP, line 45: do you prefer the RAILS_ENV and SECRET_KEY in the RUN over setting them as ENV above?

.rubocop.yml
```diff
@@ -94,6 +94,7 @@ Metrics/MethodLength:
     - 'apply'
     - 'consumer_signup'
     - 'build_new_application_payload'
+    - 'apis_list'
```
FonzMP, line 97: What about breaking the piece that structures the hash to a separate method and we wouldn't need to add this? May be a reusable method at some point as well?

app/controllers/admin/apis_controller.rb
```diff
@@ -2,22 +2,11 @@
...
-
-  def create
-    api = manage_api api_params
-    if api.save
-      render json: ApiSerializer.render(api)
-    else
-      render json: { error: api.errors.full_messages }, status: :unprocessable_entity
-    end
-  end
-
```
FonzMP, line 16: did we remove the create entirely or am I overlooking a move? Is the thought that with the bulk upload, we won't need a create? or did this not get relocated?

Dockerfile
SlakrHakr: yeah, makes it more explicit that the environment variables are needed just for this run command
i initially took them out because i didnt think they should be needed to build the container. it makes sense why theyre needed here now, but i think it would have been more explicit this way imo

app/controllers/admin/apis_controller.rb
SlakrHakr: this was removed. my understand is that this isn't something we're really using right now.
as i understand it, all we're really using right now is the dashboard to populate the environment and the apply endpoint that developer-portal submits to. 
I wanted to remove everything we weren't immediately using to get us back down to bare bones and build on from there

SlakrHakr: that being said, is this endpoint something we need in place right now?

.rubocop.yml
SlakrHakr: https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/48/commits/5c314a81b9d4b494ce817ef78ef37f71c5060c89

FonzMP (approved):
:shipit:

PR #60, opened 2022-02-21
Api 12620 production access

FonzMP:
Sends the production access request emails via LPB. The email templates and styling were brought over from the dev portal backend.
The dev environment provides previews at `/rails/mailers` and also writes to `tmp/mail`.

Also updates gems to pass brakeman.


app/api/validators/length.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Length < Grape::Validations::Validators::Base
+  def validate_param!(attr_name, params)
+    unless params[attr_name].length <= @option
```
SlakrHakr, line 5: should this maybe check for nil here?

app/javascript/packs/application.js
```diff
@@ -13,5 +13,3 @@ import "channels"
 Rails.start()
 Turbolinks.start()
 ActiveStorage.start()
-
-console.log('hello')
```
SlakrHakr, line 17: 😂

FonzMP: yeah. I figured this was safe to remove. Hopefully it doesn't take out our lower environments. 😆

lib/config_helper.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module ConfigHelper
+  module_function
+
+  def setup_action_mailer(config)
```
SlakrHakr, line 6: nit: this being broken out like this feels a bit odd to me, just because this should only be happening when LPB is run in the "production" environment. 
this being broken out here makes me think its reusable, but i dont think it really is?

spec/rails_helper.rb
```diff
@@ -48,6 +48,10 @@ def self.all_matches
   exit 1
 end
 
+SimpleCov.start do
+  add_filter 'lib/config_helper.rb'
```
SlakrHakr, line 52: this filter should be added here instead:
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/slakrhakr/API-11956/spec/spec_helper.rb#L11

lib/config_helper.rb
FonzMP: as of current, correct. The original thought was to do something like vets-api where we use it across all environments and we set the config to work appropriately based on environments and variables. In the current state I can move it just to prod if we want to remove it.

app/api/validators/length.rb
FonzMP: looks like it doesn't hit the validator if a value is not provided. The one that is actively using this is optional even. We could definitely add a check, but I think Grape is handling the dirty work here.

FonzMP: i stand corrected. there is a way to break it. to literally pass `null`

```"detail": "undefined method `length' for nil:NilClass"```

File: spec/rails_helper.rb (hunk 4 above)
Comment (FonzMP) [reply spec/rails_helper.rb:52]: actually, with config_helper gone, we should be able to eliminate this all together.

File: app/api/v0/consumers.rb [hunk 5]
```diff
@@ -74,6 +80,65 @@ def missing_oauth_params?
...
+        requires :valueProvided, type: String
+        optional :vasiSystemName, type: String
+        requires :veteranFacing, type: Boolean
+        optional :veteranFacingDescription, type: String, length: 415
+        optional :vulnerabilityManagement, type: String
+        optional :website, type: String
+      end
+
+      post 'production_request' do
+        status 200
```
SlakrHakr, line 139: it just occurred to me, if we're wanting to not send back a body, this status should probably be a 204. not sure if that would be too much of a pain to deal with within developer-portal but wanted to mention it

app/api/validators/length.rb
FonzMP: https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/60/commits/064004a823eb449607300c2b158498470cb66725

spec/rails_helper.rb
FonzMP: Removed config_helper so no longer needed.

app/api/v0/consumers.rb
FonzMP: 204 now. dev portal still works fine locally with it. https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/60/commits/bdd1675e732efdc2e6e9f145fc170b86bd1d1818 is an alias for 204. Are we fine with this, or would it make more sense to display the entirety. Would be something along the lines of 
```
body nil
status 204
```

lib/config_helper.rb
SlakrHakr: i only mentioned this initially because i couldnt think of a situation where any of the rails environments other than production should be sending out emails.
it gets confusing because im talking about rails environments, not our deployed environments, but all of our deployed environments should be running rails production environment.

SlakrHakr (approved):

PR #63, opened 2022-02-25
replace assign_attributes with update

FonzMP:
Swap `assign_attributes` to `update` within the `seeds.rb`. 

Per documentation, it looks like `assign_attributes` only applies the changes [once the object is saved](https://apidock.com/rails/ActiveRecord/AttributeAssignment/assign_attributes). 
That being said, it's only impacting the `acl` and `auth_server_access_key`, which are root level. I believe the API model is handling the creating of `api_ref`, `api_metadatum`, etc; through `find_or_create_by`. In my local (using tablePlus), all `acl` and `auth_server_access_key` are showing `NULL` on master. Switching to `update` fixes this, but I'm curious if it will have other impacts on the other changes merged today?

Another note, our dev instance is 503 cause we forgot to clear the database before running `db:migrate` for flipper, and the lpb-api-providers, so we'll have to find a way to fix that. 😭


FonzMP:
Initial build failed due to the `Api.last` item I mentioned earlier. Passed on second build, so ignore the red X above. All checks have passed is accurate.

SlakrHakr (approved):

PR #65, opened 2022-03-02
Api 13797 add emails to sandbox

FonzMP:
Adds the welcome email and va_profile email to sandbox signup.

### DIFF
--- a/app/helpers/mailer_helper.rb
+++ b/app/helpers/mailer_helper.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module MailerHelper
+  def list_apis(apis)
+    api_list = from_db
+    apis.split(',').map { |api| api_list[api] }.to_sentence
+  end
+
+  private
+
+  def from_db
+    {}.tap do |api_builder|
+      Api.kept.map { |api| api_builder[api.api_ref.name] = api.api_metadatum.display_name }
+    end
+  end
+end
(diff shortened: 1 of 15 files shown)


app/helpers/mailer_helper.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module MailerHelper
+  def list_apis(apis)
+    api_list = from_db
+    apis.split(',').map { |api| api_list[api] }.to_sentence
```
SlakrHakr, line 6: to_sentence is something im gonna have to remember
one of the reasons i really enjoy ruby, random fancy methods like that 👍🏻

spec/factories/api_metadata.rb
```diff
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :api_metadatum do
     description { Faker::Lorem.sentence }
-    display_name { Faker::JapaneseMedia::Naruto.character }
+    display_name { "#{Faker::JapaneseMedia::Naruto.character} API" }
```
SlakrHakr, line 6: is there a reason an API display name has to end in API ?

SlakrHakr (approved):

spec/factories/api_metadata.rb
FonzMP: not necessarily. glad to revert it back if you have a preference, but the display name typically ends with `API` and for testing purposes and the mailer previews, I wanted to add it for clarity. Using Naruto names looks interesting. (ex: `APIs: Buna and F` - as mentioned, not necessary, but it looks cleaner.

PR #66, opened 2022-03-02
use swagger-ui as admin dashboard

SlakrHakr:
Replaces current admin/dashboard page with full list of endpoints and moves current admin dashboard functionality into simple endpoints

Benefit: Gives us access to all endpoints we create for free (since hitting APIs on CAG is a bit of a pain)

image


config/initializers/rails_admin.rb
```diff
@@ -4,16 +4,10 @@
...
-
-  ## == CancanCan ==
-  # config.authorize_with :cancancan
-
-  ## == Pundit ==
-  # config.authorize_with :pundit
+  config.authenticate_with do
+    warden.authenticate!(scope: :user) if Figaro.env.enable_github_auth.present?
+  end
+  config.current_user_method(&:current_user)
```
FonzMP, line 10: any particular reason we added this back in? Assuming just so we can toggle it if needed? Or limit access? But just interested in the thought process.

config/routes.rb
```diff
@@ -9,34 +9,15 @@
     mount OkComputer::Engine, at: '/health_check'
     devise_for :users, controllers: { omniauth_callbacks: 'users/omniauth_callbacks' }
 
-    mount RailsAdmin::Engine => '/admin/rails_admin', as: 'rails_admin'
+    mount RailsAdmin::Engine => '/admin/database', as: 'rails_admin'
```
FonzMP, line 12: do we want to add some kind of auth to this as well? then it's all in the same category for when / if needed? also not necessary for this PR, just thinking out loud.

FonzMP (approved):
I'm good with this overall. I can't really think of a reason that we wouldn't  want to do this. A few questions, but nothing mandating changes, just questions / thoughts in general.

config/routes.rb
SlakrHakr: thats what https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/66/files#diff-b2b902da0e7e8973e6ae3bbfba8063c408080c37d7fb608c8bfee6b7ebf6a817 is doing, adding auth if toggled on through an environment variable

config/initializers/rails_admin.rb
SlakrHakr: this is adding auth to the rails_admin page (if toggled on through the environment variable)

FonzMP (approved):
:shipit:

PR #70, opened 2022-03-08
Api 7686 usage statistics

FonzMP:
Provides the ability to query ES for a successful first call. To test this locally, you'll need to generate a User and Consumer within local. The consumer should have either one or both of the `jwt_claims.cid` or `consumer.id` from the `spec/support/elasticsearch/mock_logs.json` as their `sandbox_(oauth/gateway)_ref`.

Thinking this through, I just realized I'll need to add the ability to select an environment, or potentially just provide both?


app/services/elasticsearch_service.rb
```diff
@@ -6,18 +6,39 @@ class ElasticsearchService
...
-    req = Net::HTTP::Post.new(uri, 'Content-Type' => 'application/json')
+    req = Net::HTTP::Post.new(uri)
     req.body = File.read(File.join(File.dirname(__FILE__), '../../spec/support/elasticsearch/mock_logs.json'))
     request(req, uri)
   end
 
+  def first_successful_call(consumer)
+    req = Net::HTTP::Get.new(@uri)
+    kong_id = consumer[:sandbox_gateway_ref]
+    cid = consumer[:sandbox_oauth_ref]
```
FonzMP, line 27: the Jira item says to just hit sandbox. Are we concerned with different env's now? Assuming we'll just enhance it's functionality as we continue working on LPB (LCMS) so this would suffice for now.

app/services/elasticsearch_service.rb
```diff
@@ -27,4 +48,59 @@ def request(req, uri)
...
+    }
+  end
+
+  def locate_ids(ids)
+    {
+      query_string: {
+        query: ids
+      }
+    }
+  end
```
FonzMP, line 105: I was thinking future-proof and tried to provide a way to 'build' queries for future work. I'm wondering if the query items are worthy of their own service some form of helper? At the current state maybe it makes sense to leave it here, and we just move once we start expanding?

app/api/v0/reports.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module V0
+  class Reports < V0::Base
+    version 'v0'
+
+    resource 'reports' do
+      desc 'Peruses Elasticsearch for a successful consumer first-call (via oauth and/or key-auth)'
+      get '/first-call/:id' do
```
SlakrHakr, line 9: nit: can we move this under consumers and give it a name thats a bit more general that can be extended? maybe something like:
/platform-backend/v0/consumers/:consumerId/statistics

app/api/v0/entities/report_entity.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module V0
+  module Entities
+    class ReportEntity < Grape::Entity
+      expose :first_call
```
SlakrHakr, line 6: nit: this endpoint is going to grow to add things like: "first call to production"
is there a better name for this value? maybe "firstSandboxInteractionAt" ?

app/services/elasticsearch_service.rb
SlakrHakr: we will want to have the same sort of timestamp for production in the future, but we can definitely enhance this logic when we get there. i think this is fine for now

app/services/elasticsearch_service.rb
SlakrHakr: i do like the idea of an "ElasticsearchQueryBuilder" service
that being said, i think what youve got here is definitely good for now, we can move it out and expand when more ES queries come up

SlakrHakr (dismissed):

app/services/elasticsearch_service.rb
```diff
@@ -6,18 +6,39 @@ class ElasticsearchService
...
+  def first_successful_call(consumer)
+    req = Net::HTTP::Get.new(@uri)
+    kong_id = consumer[:sandbox_gateway_ref]
+    cid = consumer[:sandbox_oauth_ref]
+    first_success_query(kong_id, cid)
+    req.body = @query.to_json
+    response = request(req, @uri)
+    if response['hits']['total']['value'].positive?
+      first_call = parse_times(response)
+      { firstSandboxInteractionAt: convert_time(first_call) }
```
SlakrHakr, line 33: ah, it just occurred to me, this is very much a nitpick but id probably name this "first_sandbox_interaction_at" here to go along with ruby convention, and then convert it to camel case in the response entity, something like this: "expose :first_sandbox_interaction_at, as: :firstSandboxInteractionAt"

FonzMP: yeah, I was interested in the camelcase vs snakecase when it was mentioned, but I like the conversion here. Adding now.

FonzMP: 58073d5

SlakrHakr (approved):

PR #75, opened 2022-03-23
API-14243: Only allow requests to developer-portal-backend from developer-portal frontend

SlakrHakr:
Flow is essentially:

developer-portal stashes a random value into a session cookie and also sends that value in the header of the request to LPB.
LPB validates that what is stored in the session cookie matches the value sent in the header.

This will require both the developer-portal and LPB to live under the same domain (in order to share cookie access).


app/api/base.rb
```diff
@@ -3,20 +3,28 @@
 class Base < Grape::API
   format :json
   rescue_from Grape::Exceptions::ValidationErrors do |e|
-    error!({ errors: [Entities::ErrorEntity.represent({ title: 'Bad Request',
-                                                        detail: e.message })] }, 400)
+    error!({ errors: e.message.split(',') }, 400)
```
SlakrHakr, line 6: Converted these errors to a list of strings to match what the developer-portal is currently expecting (so the portal can render it appropriately)

app/api/base.rb
```diff
@@ -3,20 +3,28 @@
...
   end
   rescue_from :all do |e|
-    error!({ errors: [Entities::ErrorEntity.represent({ title: 'Internal Server Error',
-                                                        detail: e.message })] }, 500)
+    error!({ errors: [e.message] }, 500)
+  end
+
+  helpers do
+    def protect_from_forgery
+      return unless Flipper.enabled? :protect_from_forgery
```
SlakrHakr, line 23: Added a flipper that defaults to off in order to not break the consumer apply endpoint for our ongoing wasa

app/api/v0/consumers.rb
```diff
@@ -62,17 +62,17 @@ def send_sandbox_welcome_emails(request, kong_consumer, okta_consumer)
     end
 
     resource 'consumers' do
-      desc 'Accept form submission from developer-portal'
+      desc 'Accept form submission from developer-portal', deprecated: true
```
SlakrHakr, line 65: I went ahead and marked all endpoints that are going to use protect_from_forgery as deprecated so it will hopefully be obvious which ones the portal (and only the portal) will use

app/api/v0/consumers.rb
```diff
@@ -62,17 +62,17 @@ def send_sandbox_welcome_emails(request, kong_consumer, okta_consumer)
...
       params do
         requires :apis, type: String, allow_blank: false
-        requires :description, type: String
+        optional :description, type: String
```
SlakrHakr, line 68: Description is optional within the portal form

app/api/v0/consumers.rb
```diff
@@ -62,17 +62,17 @@ def send_sandbox_welcome_emails(request, kong_consumer, okta_consumer)
...
         requires :firstName, type: String
         requires :lastName, type: String
-        optional :oAuthApplicationType, type: String, values: %w[web native], allow_blank: false
+        optional :oAuthApplicationType, type: String, values: %w[web native], allow_blank: true
         optional :oAuthRedirectURI, type: String,
-                                    allow_blank: false,
-                                    regexp: %r{^https?://.+},
+                                    allow_blank: true,
+                                    regexp: %r{^(https?://.+|)$},
                                     malicious_url_protection: true
```
SlakrHakr, line 76: These keys always get sent, they just have an empty string if none of the APIs being applied for implement oauth

spec/factories/api_refs.rb
```diff
@@ -2,6 +2,6 @@
 
 FactoryBot.define do
   factory :api_ref do
-    name { Faker::Hipster.word }
+    name { Faker::Hipster.unique.word }
```
SlakrHakr, line 5: Noticed a sporadic failed test if this factory happens to create duplicate api ref names

gileswells:
I wonder if matching a header to a cookie is good enough for this use case. I mean how aggressively are we wanting to enforce the same domain requirement? Should we also be leveraging referer for this? Is CORS an option here?

SlakrHakr:
> I wonder if matching a header to a cookie is good enough for this use case. I mean how aggressively are we wanting to enforce the same domain requirement? Should we also be leveraging referer for this? Is CORS an option here?

I initially went down the route of checking the referer but that seemed incredibly easy to fake from what I could tell?

I went away from CORS mostly because it doesn't seem to stop curl (or postman, etc.) requests.

That being said, having both this cookie implementation alongside CORS? that could make sense. I'll add a CORS header to this PR.

app/api/v0/consumers.rb
FonzMP: wonder if we can write our own custom grape item here instead of using deprecated? fine with utilizing deprecated currently, but think it's worth looking at in the future. I'm just afraid this may add some confusion.

SlakrHakr: The reason I went with deprecated here is because
1. It displays differently in swagger-ui so, hopefully, obvious which endpoints cant be used directly in swagger-ui
2. These endpoints are designed to work with no changes to the frontend. I could benefits to potentially improving them to be a bit more RESTful in the future but that would require changes to the frontend as well so 🤷🏻 

With regards to #1 though. We alternatively could just hide these endpoints from swagger-ui entirely? I do like the idea of being able to interact with them if need be though.

app/api/base.rb
SlakrHakr: Related to my comment below, it could be useful to quickly be able to hit these endpoints directly for the time being. I do think it makes sense to eventually remove it though, especially when LCMS is the developer-portal's official backend, but it seems worth keeping in for now i think

PR #76, opened 2022-03-25
Upgrade to rails 7

SlakrHakr:
Rails 7 was blocked for us due to the rails_admin gem. The rails_admin was just updated to 3.0.0 and it supports rails 7 now.


FonzMP:
somehow the `DateTime.now` we were using is being converted within the model to `ActiveSupport::TimeWithZone`. The real issue comes from the rounding it's doing on the last four digits within the model. Trying to find a recent change that would impact this in the Rails repo, but no luck as of yet. I switched `DateTime.now` to `Time.zone.now` which made it the correct class, but it was still rounding the last 4 milliseconds to 0's. To alleviate this, I just change all of the milliseconds to 0's upon creation with `Time.zone.now.change(:usec => 0)`.

FonzMP (approved):
:shipit:

PR #77, opened 2022-03-25
flexible docker-compose configuration

SlakrHakr:
After setting up a new computer, I made a few docker-compose configuration changes that
1. Help me get this project running again locally and
2. Should be more flexible for future maintainers


FonzMP:
I'm torn on this PR. 
I like the idea of adding the ports cause it is kind of annoying to have to add them in manually each time. With that being said, breaking out the dependencies to a separate docker file seems redundant to what we already have. Using override is nice, because docker traverses the directory looking for the `override.yml` file. Could we not add a `.ports.yml` file and just `-f docker-compose.ports.yml` ?     Of course, all this stems from the `port already in use` error we see on occasion after a failed build. I feel like correcting that would be helpful - glad to look in to that a bit. If we did this, we could just add the ports directly to the `override.yml`.

Another item is that I'm not sure moving the DB makes a ton of sense to me as you can run the application locally without it and it won't be impacted. Currently I'm able to `rails s` for any application development, and when I need the dependencies I have been adding the ports to each of the dependencies and running `docker-compose -f docker-compose.override.yml up -d`. We can't simply add ports to the `override.yml` due to the `port already in use` issue we see on occasion, but maybe adding a `ports.yml` file would be better to simplify the aforementioned process? 

Personally I'm just thinking that this is overcomplicating the process. We've mounted the DB, but if we want to work on the application, we're going to have to mount the application itself as well for this to fully run in docker. Currently, changes within  the application won't display without a `down` and `up`. 

```
volumes: 
  - .:/home/ruby
  ```

I believe I tried this before and got a gemfile issue in Jenkins, but may be worth looking back in to. Getting this entirely in docker would make sense - but I'm just trying to think of the best approach.

SlakrHakr:
Fair enough. These changes stemmed from me getting this going on my new machine. I was shooting for not installed postgres natively, hence why i moved the db container to the dependencies file.

The 3 situations I was trying to accommodate here is:
1. CI needing to run all containers without ports
2. Local developer wanting to run all containers with ports
3. Local developer wanting to run all dependency containers with ports, but run rails natively (me)

I can definitely understand the dislike of the added complexity though. I'll keep this one around for a little bit, maybe I can come up with something that works for the above scenarios but also isnt quite as complicated as this PR currently is

FonzMP:
ok, the db move makes sense now. I can see putting that in to dependencies. I didn't think through the postgres install. 

I was messing with a docker-compose rework on Friday and what I ended up with essentially just swapped all the `localhost:9200` to `service:port` and mounted the application. Once we do that, we can hit the all the services using that service names as long as we're running it through docker. The only addition I was thinking of adding was the `ports.yml` similar to what you have and then you could access them via localhost if you run `-f docker-compose.port.yml`

The risk here was that I did have to bind the port for the application to be able to see it locally, so it does have one `ports:` section for the application, which would run the risk of failing with the error we previously saw. With it being only one potential port conflict, I feel like it's much less of a risk, but still a risk.

Overall it wasn't much different in respect to this, but you would be able to just `docker-compose up` to have all the application and dependencies up and running.

FonzMP:
yeah, this is the one that got me.

> Local developer wanting to run all dependency containers with ports, but run rails natively (me)

I run the `override.yml` currently to run all external dependencies, and then just `rails s`, I just spaced that I have postgres installed 😆

SlakrHakr:
The more I think about this the more I think it could be valuable.
Build should be passing now, if we decide to move forward with it.

FonzMP (dismissed):
:shipit:

FonzMP (approved):
🚢

PR #78, opened 2022-03-30
Api 7675 promote to prod

FonzMP:
Tooling to promote a consumer to the production environment via a list of Api Refs (similar to the signup/apply page). Validates the consumer has access to the APIs within sandbox before sending off the kong / okta interactions, then updates the respective consumer fields upon success. If the consumer does not have access to the APIs within sandbox, will return an `unprocessable entity` response with a generic warning.

Added a way to retrieve a list of APIs per environment for a given consumer as well. This should provide salesforce a way to populate a list for the promotion tooling. Initially looked at using `sandbox` or `production` vs `:environmentId`, but to keep it restful elected to go with `:environmentId`. That being said, wanted to open an `environments/` endpoint to display all available environments to easily locate the ID.


FonzMP:
Thinking through this process earlier. Wanted to jot down some (what I believe to be) mentionable items.

1. We're only using the email to find a user / consumer and we're overwriting the current data. This includes any current `_refs`. I'm curious how we can utilize the existence of a current `_ref` to check for an Okta Application or Kong Consumer and append the necessary info, therefore keeping the same user / consumer details. That being said, this could potentially be a risk, which leads to the below.
2. How do we validate that a user is signing up with the correct info? Until we have user accounts, anyone could use an email to sign up and by us trusting this as the identifier for a user / consumer within LCMS, we run the risk of potentially displaying information that we shouldn't. 


Seems like a fine line in it's current state. I'm sure there's much more to discuss here, whether it be this being the best approach until User Accounts (which may or may not be the case), or other potential resolutions, but just wanted to note these.

FonzMP:
I'm seeing the `/utilities/consumers` response and I'm starting to think we can just utilize that for the APIs and Environments.  Not sure how I didn't notice it before, but may be worth trimming off the excess here as you've already done the heavy lifting. @SlakrHakr

app/api/v0/consumers.rb
```diff
@@ -175,6 +200,36 @@ def send_sandbox_welcome_emails(request, kong_consumer, okta_consumer)
...
+        end
+        consumer = Consumer.find(params[:consumerId])
+        consumer_api_refs = consumer.apis.map(&:api_ref).map(&:name)
+
+        api_refs = validate_refs(consumer_api_refs)
+        if api_refs.empty?
+          status 422
+          response = { title: 'Invalid API Refs',
+                       detail: "This consumer is not approved for #{params[:apis].split(',').to_sentence} in Sandbox" }
+          present response, with: V0::Entities::InvalidRequestEntity
```
SlakrHakr, line 223: We should handle errors by raising an exception and managing those exceptions in the base file. example: https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/master/app/api/base.rb#L14

app/api/v0/consumers.rb
```diff
@@ -175,6 +200,36 @@ def send_sandbox_welcome_emails(request, kong_consumer, okta_consumer)
         first_call = ElasticsearchService.new.first_successful_call consumer
         present first_call, with: V0::Entities::ConsumerStatisticEntity
       end
+
+      desc 'Promotes a consumer to the production environment for the provided API(s)'
+      post '/:consumerId/promote' do
```
SlakrHakr, line 205: nit: custom verbs are not supposed to be in the url of a rest api. we could be consistent with the rest of the api by changing this to be `/promotion-requests` as a post to that named endpoint would be "creating and appending a promotion request"

app/api/v0/consumers.rb
```diff
@@ -22,19 +22,28 @@ def user_from_signup_params
...
+      def okta_signup(user, oauth, env = nil)
+        okta_consumer = OktaService.new(env).consumer_signup(user,
+                                                             oauth,
+                                                             application_type: params[:oAuthApplicationType],
+                                                             redirect_uri: params[:oAuthRedirectURI])
+        if env.nil?
+          user.consumer.sandbox_oauth_ref = okta_consumer.id
+        else
+          user.consumer.prod_oauth_ref = okta_consumer.id
+        end
```
SlakrHakr, line 46: nit: this is much more just a preference thing, but the more ive written ruby (and dealt with rubocop) the more ive tried to work out blocks like this, typically replacing with one or more guard clauses:
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GuardClause

so this could be changed to
```
user.consumer.sandbox_oauth_ref = okta_consumer.id if env.blank?
user.consumer.prod_oauth_ref = okta_consumer.id if env.present?
```

app/api/v0/entities/invalid_request_entity.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module V0
+  module Entities
+    class InvalidRequestEntity < Grape::Entity
+      expose :title, documentation: { type: 'String' }
+      expose :detail, documentation: { type: 'String' }
+    end
+  end
+end
```
SlakrHakr, line 10: We can bring this sort of error entity back but I would say it should be a general ErrorEntity if thats what we want to do.
I took the previous general ErrorEntity out in this commit
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/commit/6a4aff3fa76dc549d3881f99a92f9d920fba2c2e#diff-c1da07af72ca418326ffbd375dad6aa095705556c034d221e3df68043022de12
when I realized the developer-portal was expecting errors in a certain format

app/services/kong_service.rb
```diff
@@ -144,8 +141,34 @@ def handle_key_auth_flow(org, last_name, key_auth)
...
+  def set_kong_elb
+    if @env.nil?
+      Figaro.env.kong_elb || 'http://localhost:4001'
+    else
+      Figaro.env.prod_kong_elb || 'http://localhost:4003'
+    end
+  end
+
+  def set_kong_password
+    return nil unless Rails.env.production?
```
SlakrHakr, line 161: I try to steer away from checking the rails environment for stuff like this. It would be nice to be able to use a password locally if need be. Could just change this back to the check that was in there:
```
return nill if Figaro.env.kong_password.blank?
```

app/services/kong_service.rb
```diff
@@ -144,8 +141,34 @@ def handle_key_auth_flow(org, last_name, key_auth)
     [kong_id, kong_consumer_name, kong_api_key]
   end
 
+  def set_client
+    if Figaro.env.socks_host.present?
+      Net::HTTP.SOCKSProxy(Figaro.env.socks_host, 2001)
+    else
+      Net::HTTP
+    end
+  end
```
SlakrHakr, line 150: preference: this method isn't necessarily needed if its converted into an inline if statement:
```
Figaro.env.socks_host.present? ? Net::HTTP.SOCKSProxy(Figaro.env.socks_host, 2001) : Net::HTTP
```

If this method is kept for clarity, etc, id vote going the guard clause route, so:
```
def set_client
    return Net::HTTP.SOCKSProxy(Figaro.env.socks_host, 2001) if Figaro.env.socks_host.present?

    Net::HTTP
 end
```

pretty much all preference here though, so nothing I'd push on or anything

app/api/v0/consumers.rb
FonzMP: [a1fa9ba](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/a1fa9ba81fb9cc05b15f8b72573f130a2b6eb7cf) 👍

app/services/okta_service.rb
```diff
@@ -3,8 +3,11 @@
 require 'oktakit'
 
 class OktaService
```
SlakrHakr, line 5: I wonder if this could be structured any better, since the environments are strictly sandbox and production, I wonder if inheritance could be used to make any of this cleaner

[inheritance-implementation-suggestion.zip](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/files/8446325/inheritance-implementation-suggestion.zip)

So OktaService would no longer be consumed directly, the rest of the application would utilize either ProductionOktaService or SandboxOktaService.

app/services/kong_service.rb
FonzMP: [f94bf5b](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/f94bf5b611a5845bf431549dd380469195145d29)

app/api/v0/entities/invalid_request_entity.rb
FonzMP: [f94bf5b](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/f94bf5b611a5845bf431549dd380469195145d29)

app/api/v0/consumers.rb
FonzMP: personally I like this better anywho. [a837766](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/a837766944887140d9266b33a36b926775a98caf)

app/services/okta_service.rb
```diff
@@ -46,6 +49,46 @@ def consumer_name(user)
     "LPB-#{"#{user.consumer.organization}#{user.last_name}".gsub(/\W/, '')}"
   end
 
+  def set_va_redirect
+    if @env.nil?
+      Figaro.env.okta_redirect_url
+    else
+      Figaro.env.prod_okta_redirect_url
```
SlakrHakr, line 56: These blanket else statements that connect to production concern me a little.
What if a consumer sent in "sandbox" as the environment? I would expect it to connect to sandbox but this implementation it would connect to production.
The inheritance suggestion above does fix this issue as well. But if we wanted to keep this single class, it might be worth checking for specific environment values being passed in and only allowing :sandbox and :production

app/services/kong_service.rb
FonzMP: always a ternary fan. not sure why I didn't use it here tbh. [3edfcfd](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/3edfcfd3e61667d64c08d41edb91f267852e3936)

SlakrHakr:
> I'm seeing the `/utilities/consumers` response and I'm starting to think we can just utilize that for the APIs and Environments. Not sure how I didn't notice it before, but may be worth trimming off the excess here as you've already done the heavy lifting. @SlakrHakr

I will say, anything under the "utilities" resource are meant for our use only. They were initially implemented as a replacement for the UI admin dashboard we had initially.
If any of this data is needed by an outside consumer (salesforce for example) we will want to implement endpoints within the v# structure to facilitate that

app/models/consumer.rb
```diff
@@ -33,6 +33,13 @@ def api_ids
     apis.map(&:id)
   end
 
+  def promote_to_prod(api_ref)
+    api = ApiRef.find_by(name: api_ref).api
```
SlakrHakr, line 37: What happens if a record isn't found here? Should we raise an exception? Return false immediately after this line?

app/services/okta_service.rb
FonzMP: I was never sold on the `set_` methods. I definitely agree we need some layer of protection vs an 'all-or-none' value. Was definitely open to other options in regards to this.

app/services/kong_service.rb
```diff
@@ -6,13 +6,10 @@
 class KongService
```
SlakrHakr, line 6: The same approach described for the OktaService below could be used for this as well
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/files#r845542803

Could have
Kong::BaseService
Kong::SandboxService
Kong::ProductionService

where sandbox and production extended base

app/services/okta_service.rb
SlakrHakr: I wonder if the structure of this would be better using a bit of nesting, so:

Okta::BaseService
Okta::SandboxService
Okta::ProductionService

All this is just a suggestion though

app/services/kong_service.rb
FonzMP: Definitely fine with moving away from the `production?` check. I think my initial thought was the most efficient way to bypass this if it were not needed, and I assumed we'd never need it outside of the `production` build. But this makes sense, we shouldn't have the prod kong without the base kong password, so validating `kong_password` should work in the same manner. [3edfcfd](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/3edfcfd3e61667d64c08d41edb91f267852e3936)

app/api/v0/consumers.rb
FonzMP: [a1fa9ba](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/a1fa9ba81fb9cc05b15f8b72573f130a2b6eb7cf)

And this one, it's simply called `whoops`.
[7c1e959](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/7c1e959a18b3a26ba936bafd4d9af21347501eda)

app/services/okta_service.rb
FonzMP: definitely agree it could use a better overall structure, which in turn would make it a little more 'bulletproof'. I like the idea here, will do a little reading myself and start implementing this (or something similar) here shortly.

app/models/consumer.rb
FonzMP: in it's current state I don't think it can receive an invalid ApiRef name as it loops through the consumer's apis and pulls the `api_ref[:name]` from each. These are the values that get sent in if the consumer has access to the API in sandbox, and the ref is provided via the request. I don't believe there is a way for an invalid ApiRef name to be provided to the array we loop through, but I've been surprised before.

From a safety perspective for any potential future use, yeah, it makes sense to put some safety catch here. I'd bet we could add these checks in a few spots as well.

app/services/kong/base_service.rb
```diff
@@ -0,0 +1,131 @@
...
+      response.tap { |res| res['ok'] = res.is_a? Net::HTTPSuccess }
+      raise 'Failed to retrieve kong consumers list' unless response['ok']
+
+      JSON.parse(response.body) unless response.body.nil?
+    end
+
+    protected
+
+    def set_kong_elb
+      nil
```
SlakrHakr, line 124: makes more sense to me to raise an exception here instead of returning nil.
elb and password being nil will probably make consuming Kong::BaseService not function, but better if it blew up in someone's face that this is an abstract class, not to be used directly, only extended

SlakrHakr (approved):
Have one last comment but its not the end of the world.

app/services/kong/base_service.rb
SlakrHakr: Could even take it a step further and use a gem like https://github.com/openSUSE/abstract_method
Full disclosure, I've never used that gem, just always wanted to check it out

FonzMP: [b21e4e3](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/b21e4e373f71711b30f1870f91bdd16c7950cd5f)

FonzMP:
[45b1376](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/45b1376fb49e8355f44bdef73be0c4ebf05b1bc5) raises same ApiValidation Error Message
```
"errors": [
     "Invalid API list for consumer"
]
```

[bf6868f](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/bf6868f132eb0c9abf985eed00531e238e9b54b9) reworks the `base_service_spec` into `sandbox_spec` for Okta.

app/models/consumer.rb
FonzMP: [45b1376](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/78/commits/45b1376fb49e8355f44bdef73be0c4ebf05b1bc5)

SlakrHakr (approved):

PR #85, opened 2022-04-13
ignore duplicate malicious urls

SlakrHakr:
Allows re-running of this without first deleting all existing data in the database.

https://github.com/zdennis/activerecord-import/#duplicate-key-ignore


FonzMP (approved):
:shipit:

PR #86, opened 2022-04-18
decode release notes from release notes repo

SlakrHakr:
Release notes sent from release notes repo are url encoded to prevent issues with quotation marks, etc. LCMS needs to decode the release note content string before storing.


FonzMP (approved):
:shipit:

PR #88, opened 2022-04-20
more debugging

SlakrHakr:
hoping see all cookies rails knows about will help understand what is going on here


FonzMP (approved):
:shipit:

PR #93, opened 2022-04-25
Api 7677 drift script oauth

FonzMP:
Creates a daily job to check okta environments. Dependent on a feature flipper to allow us to toggle on/off easily.


app/services/okta/base_service.rb
```diff
@@ -47,10 +47,18 @@ def consumer_name(user)
       "LPB-#{"#{user.consumer.organization}#{user.last_name}".gsub(/\W/, '')}"
     end
 
+    def lower_env?
+      Figaro.env.denote_okta_lower.present?
+    end
```
FonzMP, line 52: Not sure I like this as is. Wondering if the build will break if we don't add `denote_okta_lower` to AWS for prod?

app/services/okta/base_service.rb
```diff
@@ -47,10 +47,18 @@ def consumer_name(user)
...
+
+    def construct_label(user)
+      "#{consumer_name(user)}-#{Time.now.to_i}#{lower_env? ? '-dev' : ''}"
+    end
```
FonzMP, line 56: This is the approach I think we'll have to take (or at least along these lines) to denote signups in lower environments in the sandbox Okta Org. This was an attempt to only worry about the signups coming from the developer portal to sandbox, and not getting dev / staging signups since they all share the same Okta.

Thinking through this, the alert will go off in all environments for any consumer without `-dev` so going to have to fix that... 🤔

app/services/okta/drift_service.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    def detect_unknown_entities(alert_list)
+      alert_list.filter do |app|
+        new_record? app
+      end
+    end
+
+    def alert_slack(consumer)
+      @slack_service ||= SlackService.new
```
FonzMP, line 39: added this line to allow rspec to generate a new service so I could test the method and get the respective file coverage needed.

config/sidekiq.yml
```diff
@@ -0,0 +1,4 @@
+okta_drift:
```
SlakrHakr, line 1: im assuming we can get rid of this file now?

app/services/okta/base_service.rb
SlakrHakr: Assuming "lower_env?" only returns true if in qa/dev and staging

I think this would work if the majority of the time we just leave the alert off in qa/dev and staging. We can have it on for testing the slack message system and then just keep it off in those environments the majority of the time. I think that should work since this is filtering down per the last 24 hours. Its our prod / okta sandbox that we really care about keeping in sync.

SlakrHakr (approved):

app/services/okta/base_service.rb
FonzMP: `lower_env?` works off of an ENV var, so it's dependent on adding a value. For this, I just have it as `present?`, but if we don't add `prod/env_var` in to AWS, I'm not sure the build will succeed so we may have to look at another way to address this.

The specific issue for this one is that we share our lower environment Okta across all  of the lower environments. Even if we only enable the flipper in sandbox for alerting, without appending something to notate that a signup came from a lower environment, we would receive all QA/Dev, Staging and Sandbox Okta applications in our production LCMS crosscheck with Sandbox environment. This happens because Sandbox information within LCMS prod would not have the consumer as it would live in QA/Dev or Staging.

Hopefully that makes sense?

SlakrHakr: > lower_env? works off of an ENV var

right, but im assuming we plan to have that environment variable present in only dev and staging?

> For this, I just have it as present?, but if we don't add prod/env_var in to AWS

i swear we have something similar. if we have to have the environment variable there, it can just be an empty string worst case scenario

---

the way its setup, we should have `-dev` appended to anything setup in dev and staging.
then when production checks for new records, it looks for anything without `-dev` in the name.
i think the way this is setup will work, assuming we leave notifications off for dev and staging, which i think makes sense anyways.

FonzMP: > right, but im assuming we plan to have that environment variable present in only dev and staging?

correct.

> i swear we have something similar. 

I thought so too. I can review a bit more. I did a quick peruse but didn't notice anything. I'll take some time this afternoon to review before I add these to AWS.

app/services/slack_service.rb
```diff
@@ -0,0 +1,13 @@
...
+class SlackService
+  def alert_slack(webhook, message)
+    uri = URI.parse(webhook)
+    client = Net::HTTP.new(uri.host, uri.port)
+    client.use_ssl = true
+    req = Net::HTTP::Post.new(uri.request_uri)
+    req.body = message.to_json
+    req['Content-type'] = 'application/json'
+    client.request(req)
+  end
```
SlakrHakr, line 12: does it make sense to use this existing gem?
https://github.com/slack-ruby/slack-ruby-client

PR #100, opened 2022-04-28
revert to rails 6 for govdelivery-tms gem to work

SlakrHakr:
Getting this error from deployed environment trying to send email using the govdelivery-tms gem:
```
undefined method `display_names' for #<Mail::UnstructuredField:0x000055e8c3d23230
```
Looking into the gem more, it looks like it requires activesupport 6 or less. Reverting rails from 7 to 6 to hopefully resolve the issue.


FonzMP (approved):
:shipit:

PR #113, opened 2022-05-04
Add slack on signup

FonzMP:
Adds slack message for sandbox signups.


app/services/slack_service.rb
```diff
@@ -6,6 +6,15 @@ def initialize
...
+    else
+      @client.chat_postMessage(channel: channel, attachments: message[:attachments],
+                               as_user: true)
+    end
+  end
+
+  def use_string?(message)
+    message.instance_of?(String)
   end
 end
```
SlakrHakr, line 20: this feels a little funky to me, but not the end of the world
it might make more sense to have
```
# where message is a hash with either "text" or "attachments" key
def alert_slack(channel, message)
  @client.chat_postMessage(channel: channel, message: message, as_user: true)
end
```

but then if you go down that road, what is this service abstraction really doing for us? does it make more sense just to interact with the client directly wherever we're sending the slack notification?

SlakrHakr (approved):

app/services/slack_service.rb
FonzMP: yeah, I didn't like the way this was done but I tried a few different variations with no luck.

I wanted something like 

```
@client.chat_postMessage(channel: channel, (use_string? ? { text: message } : { attachments: message[:attachments] }), as_user: true)
```
but it didn't seem to like that either. was real finicky. there is probably a better way to address this but Juice vs Squeeze. That being said, we're going to have to add the ability to use `blocks` for the Weekly report, so I'll see if I can find a better way to address this there.

PR #115, opened 2022-05-05
claims is only oauth in DPB for sandbox signup

SlakrHakr:
Email from signing up in dev-developer.va.gov
- from DPB: only provides oauth information
- from LPB: provides both api key information as well as oauth information

This should make it so signups to Benefits Claims in sandbox are only for the oauth implementation. Which makes sense considering v0 (the version that uses apikey) is deprecated and going away.


FonzMP (approved):
:shipit:

PR #120, opened 2022-05-17
Api 15713 weekly report

FonzMP:
Creates the job to display the weekly signup report. 

This runs off of the `event.created_at` so the first week will be skewed after running the import. Once those are passed, it should function correctly. Tested locally with signups by manipulating the `created_at` dates as well as using the dynamo dev table for a 'realistic' report.


app/api/utilities.rb
```diff
@@ -73,5 +73,14 @@ class Utilities < Base
         Kong::SandboxService.new.list_all_consumers
       end
     end
+
+    resource 'events' do
+      desc 'Imports events from Dynamo DB'
+      get '/' do
```
SlakrHakr, line 79: i wonder if this endpoint is needed? why not just inject this event logic into the existing consumer import logic?

if this does make sense as a separate endpoint here, its a small thing, but just probably be a `post` and renamed to be something like `event-migration-requests` to match with the above `consumer-migration-requests` and be a little clearer on what the endpoint is actually doing.

app/models/event.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class Event < ApplicationRecord
+  validates :event_type, presence: true
+  validates :event, presence: true
```
SlakrHakr, line 5: this is maybe a bit nitpicky but the record is already called `event`
id go with `type` and `content` for the columns maybe? to avoid things like: `Event.first.event`

app/models/event.rb
```diff
@@ -0,0 +1,13 @@
...
+  serialize :event
+
+  EVENT_TYPES = { sandbox_signup: 'sandbox_signup' }.freeze
+
+  def sandbox_signup_per_span(span)
+    Event.where(event_type: EVENT_TYPES[:sandbox_signup], created_at: span)
```
SlakrHakr, line 11: this might be a good spot for a `scope` instead of a method here
https://apidock.com/rails/ActiveRecord/NamedScope/ClassMethods/scope

app/services/dynamo_import_service.rb
```diff
@@ -0,0 +1,15 @@
...
+  def import_dynamo_events
+    table_data = DynamoService.new.fetch_dynamo_db
+    events = build_events(table_data['items'])
+    Event.upsert_all(events)
+  end
+
+  private
+
+  def build_events(events)
+    events.map { |event| { event_type: Event::EVENT_TYPES[:sandbox_signup], event: event } }
```
SlakrHakr, line 13: does dynamo not have some sort of timestamp within its table? id assume it would need something for the report that is run currently to work?

if it does, can we use it as the `created_at` date here for this import? to keep the report consistent (even the first week) and to keep the data as accurate as possible.

app/services/slack/report_service.rb
```diff
@@ -0,0 +1,141 @@
+# frozen_string_literal: true
+
+module Slack
+  class ReportService < AlertService
```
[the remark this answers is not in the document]
SlakrHakr:
```
select
  (case when created_at > 'week-ago-date-here' then 1 else 0 end) as weekly_count,
  count as total_count
from events
where type = 'sandbox_signup'
```

if each event has multiple apis that need to be tracked:
1. can we make each event a single api ?
2. if #1 doesnt make sense, could switch the `1` in the query above for the value of apis in the `content` field of the record, maybe...

app/api/utilities.rb
FonzMP: I think I was just trying for consistency here as we have the consumer import within the utilities.  That being said, doing it as a post (like the consumer import) does makes more sense. 👍

app/models/event.rb
FonzMP: noted. 👍

app/models/event.rb
FonzMP: noted. 👍

app/services/dynamo_import_service.rb
FonzMP: I'll look at this. Importing with a set `created_at` never even crossed my mind. Will dive in to this.

app/api/utilities.rb
FonzMP: added. [504045d](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/120/commits/504045d097eccd46af42ed64a7904c6d49b2c6fa) and [ab32d17](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/120/commits/ab32d17ba3ab85f7421e9e6e8f647856c2d463bb)

app/models/event.rb
FonzMP: swapped to content. type would override a default which I'm not real keen on. [ab32d17](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/120/commits/ab32d17ba3ab85f7421e9e6e8f647856c2d463bb)

app/models/event.rb
FonzMP: swapped with event structure. [ab32d17](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/120/commits/ab32d17ba3ab85f7421e9e6e8f647856c2d463bb)

app/services/dynamo_import_service.rb
FonzMP: Created at added. [3d4faa6](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/120/commits/3d4faa68581b53f54497d71f322fb5acdfc9a354)

app/services/slack/report_service.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    def generate_query
+      sort_refs.map do |ref|
+        "#{new_query(ref)} #{all_time_query(ref)}"
+      end
+    end
+
+    def new_query(ref)
+      "SUM((case WHEN (created_at > '#{1.week.ago}' AND content->'email' NOT IN (SELECT DISTINCT(content->'email') FROM events WHERE created_at < '#{1.week.ago}') AND content->>'apis' LIKE '%#{ref}%') then 1 else 0 end)) as #{ref}_new_count,"
```
SlakrHakr, line 23: should we do a straight = instead of LIKE when checking the ref? i worry that you might have one api named `claims` and another named `claimsBenefit` which could throw these numbers off, but maybe that wouldnt happen?

app/services/slack/report_service.rb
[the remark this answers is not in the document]
SlakrHakr: nice! love the new single query here

SlakrHakr (approved):

app/services/slack/report_service.rb
SlakrHakr: i wonder if you could use `string_to_array` (or something like that) to split the string and then call `in` on that, instead of = or LIKE

SlakrHakr (approved):

PR #121, opened 2022-05-17
Update seed service

FonzMP:
Moves dynamo seeds into the seed service and updates rake tasks to be consistent - updated README to match.


README.md
```diff
@@ -42,10 +42,8 @@ export RAILS_ENV=development
 rake db:create
 rake db:migrate
 rake db:seed
-rake dynamo:migrate
```
SlakrHakr, line 45: is this not still needed?

SlakrHakr (approved):

README.md
FonzMP: it would not be... the new [seed](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/4cc4ae15fd56e8702cdb4263e1c7dbd562c159af/rakelib/load_dynamo.rake#L19) rake task would run [seed_dynamo](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/4cc4ae15fd56e8702cdb4263e1c7dbd562c159af/app/services/utility/seed_service.rb#L76) which initializes and seeds. Migrate just initializes.

PR #124, opened 2022-05-19
API-15236: [LPB] Update sandbox apply endpoint to handle CCG auth api requests

SlakrHakr:
Adds the ability to sign up for an API using the Client Credentials Grant oauth flow.
Also includes a bit of restructure, primarily due to the need to create another okta application and track it separately.


app/services/okta/drift_service.rb
```diff
@@ -5,18 +5,14 @@ class DriftService
...
       applications = @client.new.list_applications
       apps_last_day = filter_last_day(applications)
       alert_on_list = detect_unknown_entities(apps_last_day) unless apps_last_day.empty?
-      if alert_on_list.present?
-        @slack_service = SlackService.new
-        alert_on_list.map do |consumer|
-          alert_slack consumer
-        end
-      end
+      alert_on_list.map { |consumer| alert_slack consumer } if alert_on_list.present?
```
SlakrHakr, line 15: changes in this file were just to get code coverage back up to 90 across the board, no functionality should have changed

app/services/okta/base_service.rb
```diff
@@ -79,37 +143,41 @@ def build_new_application_payload(user, application_type:, redirect_uri:)
...
-      raise 'NotImplemented'
-    end
-
-    def okta_token
-      raise 'NotImplemented'
-    end
-
-    def va_redirect
-      raise 'NotImplemented'
+    def build_new_ccg_application_payload(user, options)
```
FonzMP, line 157: Are these (`ccg_app_payload` vs `acg_app_payload`) complex enough we couldn't work them in to the same? Seems like that maaaayyyyy be the case, just curious. Assuming that's why they're split.

app/api/entities/consumer_entity.rb
```diff
@@ -6,9 +6,7 @@ class ConsumerEntity < Grape::Entity
     expose :description
     expose :tos_accepted_at, as: :tosAcceptedAt
     expose :tos_version, as: :tosVersion
-    expose :sandbox_gateway_ref, as: :sandboxGatewayRef
-    expose :prod_gateway_ref, as: :prodGatewayRef
-    expose :prod_oauth_ref, as: :prodOauthRef
```
FonzMP, line 11: did we not have sandbox_oauth? 😆

app/services/okta/production_service.rb
```diff
@@ -19,5 +19,9 @@ def okta_token
     def va_redirect
       Figaro.env.prod_okta_redirect_url
     end
+
+    def environment_key
+      'prod'
+    end
```
FonzMP, line 25: should we make these `production` for consistency (match environments in LPB / Dev Portal)? not a change that we really HAVE to do but does it make sense? It's a bit longer but I'm kind of leaning `production` vs `prod`.

app/services/okta/base_service.rb
SlakrHakr: i would say yes, a few minor things might be able to be shared between them, but theyre pretty different concepts within that method's logic, id worry sharing anything might cause problem / be confusing in the future

FonzMP (dismissed):
:shipit:

FonzMP (approved):
:shipit:

FonzMP (approved):
:shipit:

PR #125, opened 2022-05-26
API-7678:  Drift Scripts for Kong LCMS

TS000:
Jira Ticket here: https://vajira.max.gov/browse/API-7678

Kong check for consumer drifts.


app/services/kong/drift_service.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'date'
+
+module Kong
+  class DriftService
+    def initialize(environment = nil)
+      @env = environment
+      @client = Kong::SandboxService
```
SlakrHakr, line 9: youll want to setup the picking of the environment by service like the okta drift service has
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/ce8f665d864be0569e701790e0d5fd12c6e449c6/app/services/okta/drift_service.rb#L7

app/services/kong/drift_service.rb
```diff
@@ -0,0 +1,55 @@
...
+    def alert_slack(consumer)
+      @slack_service ||= SlackService.new
+      message = build_message(consumer)
+      @slack_service.alert_slack(Figaro.env.slack_drift_channel, message)
+    end
+
+    def new_record?(consumer)
+      cid = consumer['created_at']
+      if cid.present?
+        Consumer.find_by(user_id: cid)
```
SlakrHakr, line 37: user_id here is our database record identifier. im guessing youd need to match `cid` here to either `sandbox_gateway_ref` or `prod_gateway_ref`

app/services/kong/drift_service.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def alert_slack(consumer)
+      @slack_service ||= SlackService.new
+      message = build_message(consumer)
+      @slack_service.alert_slack(Figaro.env.slack_drift_channel, message)
+    end
+
+    def new_record?(consumer)
+      cid = consumer['created_at']
```
SlakrHakr, line 35: wait, i missed this, cid is a timestamp of when the kong consumer was created...? i assumed it was the kong identifier, the kong id is what you would need to match on here, unless im missing something

app/services/kong/drift_service.rb
```diff
@@ -0,0 +1,55 @@
...
+      @client.new.list_all_consumers
+    end
+
+    def time_sort
+      pull_kong_consumers.filter do |consumer|
+        consumer if consumer['created_at'] >= Date.yesterday.to_time.to_i
+      end
+    end
+
+    def detect_unknown_entities(alert_list)
```
SlakrHakr, line 22: is this whats actually going to be executed from the background scheduler? maybe have a usage example would help here

app/services/kong/drift_service.rb
```diff
@@ -0,0 +1,55 @@
...
+      alert_list.filter do |cid|
+        new_record? cid
+      end
+    end
+
+    def alert_slack(consumer)
+      @slack_service ||= SlackService.new
```
FonzMP, line 29: this slack service has changed now as well. You'll want to use the `AlertService` to `send_message`.

config/database.yml
```diff
@@ -3,9 +3,9 @@ default: &default
   encoding: utf8
   pool: 5
   timeout: 5000
-  host: <%= ENV['DATABASE_HOST'] %>
-  username: <%= ENV['DATABASE_USER'] %>
-  password: <%= ENV['DATABASE_PASSWORD'] %>
+  host: localhost
+  username: postgres
+  password: postgres
```
FonzMP, line 8: these shouldn't be changed. This is set by running the values locally or dependent on the environment variable.

SlakrHakr, line 8: we want to make sure to not commit this

app/services/kong/drift_service.rb
```diff
@@ -0,0 +1,55 @@
...
+        Consumer.find_by(user_id: cid)
+      end
+    end
+
+    def trim_url(url)
+      url[0...-5]
+    end
+
+    def build_message(consumer)
+      url = trim_url(consumer[:_links][:uploadLogo][:href]) if consumer[:_links].present?
```
SlakrHakr, line 46: pretty sure this uploadLogo bit is okta specific?

app/services/kong/drift_service.rb
```diff
@@ -0,0 +1,55 @@
...
+      environment = production? ? 'Production' : 'Sandbox'
+      [
+        '*Lighthouse Consumer Management Service Notification*',
+        "Detected an unknown Consumer within Kong Environment: #{environment}",
+        "Client ID: <#{url}|#{consumer[:credentials][:oauthClient][:client_id]}>"
```
SlakrHakr, line 51: this mentions oauthClient, im assuming this doesnt work?

app/services/kong/drift_service.rb
```diff
@@ -0,0 +1,55 @@
...
+    def new_record?(consumer)
+      cid = consumer['created_at']
+      if cid.present?
+        Consumer.find_by(user_id: cid)
+      end
+    end
+
+    def trim_url(url)
+      url[0...-5]
+    end
```
SlakrHakr, line 43: this isnt needed since the usage below is okta specific

config/database.yml
TS000: Ah shoot I thought I left that out of my commit. I'll remove it.

app/services/kong/drift_service.rb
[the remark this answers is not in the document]
TS000: Good catch

FonzMP:
> I don't understand the reason for the CI fail. :(

this is caused due to commiting the database connection info. 
```
  host: localhost
  username: postgres
  password: postgres
  ```
  
  it's trying to connect to localhost within the container where as we want it to connect to the service.
  https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/ce8f665d864be0569e701790e0d5fd12c6e449c6/docker-compose.yml#L20
we utilize the docker compose to set the container variables.

File: app/services/kong/drift_service.rb [hunk 10]
```diff
@@ -0,0 +1,66 @@
...
+      @env = environment
+      @client = production? ? Kong::ProductionService : Kong::SandboxService
+      @slack_service = Slack::AlertService.new
+    end
+
+    def detect_drift
+      consumers = pull_kong_consumers
+      consumers_last_day = filter_last_day(consumers)
+      alert_on_list = detect_unknown_entities(consumers_last_day) unless consumers_last_day.empty?
+      alert_on_list.map { |consumer| alert_slack consumer } if alert_on_list.present?
```
SlakrHakr, line 15: could use `each` instead of `map` here
both iterate over the collection but map returns a new collection with each item modified

SlakrHakr (dismissed):

SlakrHakr (approved):

PR #138, opened 2022-06-02
client id in response is not what developer-portal expects

SlakrHakr:
This issue was causing the developer-portal frontend to not display oauth information on the sandbox signup success page.


FonzMP (approved):
good catch.
:shipit:

PR #139, opened 2022-06-03
clean up utility endpoints, remove hard delete actions

SlakrHakr:


app/api/utilities.rb
```diff
@@ -36,20 +29,6 @@ class Utilities < Base
...
-                           allow_blank: false,
-                           default: false,
-                           values: [true, false],
-                           description: 'False simply flags the records in the db, true destroys them forever'
-      end
-      delete '/' do
-        params[:destroy] ? User.destroy_all : User.discard_all
-
-        { success: true }
-      end
```
SlakrHakr, line 51: Now that we're officially deployed in production, i wanted to remove the outright full deletes from this dashboard, to prevent accidental submissions.

app/models/api_category.rb
```diff
@@ -3,5 +3,7 @@
 class ApiCategory < ApplicationRecord
   include Discard::Model
 
+  validates :key, uniqueness: true
```
SlakrHakr, line 6: had a strange issue locally where i had duplicate categories that had me stumped for a bit, fixed my local db and put this in place so it doesnt happen in the future

lib/tasks/data_migrations.rake
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+namespace :data_migrations do
```
SlakrHakr, line 3: wanted a nice way to add/modify apis through each environment in a consistent way (at least while we're maintaining it somewhat manually for now).

typically we would just create these rake tasks and run them in the deployed environment.
since we dont have easy access to the terminal in our deployed environments, anything under this namespace should be executable from the "data-migrations" endpoint on the dashboard

app/models/api.rb
```diff
@@ -29,8 +29,9 @@ def status
   end
 
   def activate!
-    self.deactivation_info = nil
-    save!
+    metadatum = api_metadatum
+    metadatum.deactivation_info = nil
+    metadatum.save!
```
SlakrHakr, line 34: found an issue after having to add code coverage for this file for whatever reason

FonzMP (approved):
:shipit:

PR #140, opened 2022-06-06
fix report numbers

SlakrHakr:


db/seeds.rb
```diff
@@ -540,7 +540,7 @@
    }
   },
   api_ref_attributes: {
-   name: 'health'
+   name: 'clinicalHealth'
```
SlakrHakr, line 543: ok, this has been changed in all deployed environments as well

FonzMP (approved):
:shipit:

PR #141, opened 2022-06-06
add provider directory api via data migrations concept

SlakrHakr:
uses https://github.com/OffgridElectric/rails-data-migrations to manage data migrations the same way rails manages database schema migrations.

super cool concept that could be nice for managing large data changes across environments consistently.

this change adds the new provider directory api as its now in both dev and staging, and could go to production any day now.


SlakrHakr:
thinking through this a bit more, im not sure this super cool gem will work for this type of situation. we very possibly could run into times where we only want to run one of the data migrations, not all "un-run" ones.

PR #143, opened 2022-06-07
Api 16385 ccg oauth

FonzMP:


app/api/base.rb
```diff
@@ -20,6 +20,10 @@ class Base < Grape::API
     Rails.logger.warn "??? #{e.message}"
     error!({ errors: ['Access is forbidden'] }, 403)
   end
+  rescue_from AuthorizationError do |e|
+    Rails.logger.warn "??? #{e.message}"
+    error!({ errors: ['Unauthorized'] }, 401)
+  end
```
FonzMP, line 26: Assuming we can just return an `:unauthorized` and remove this and the `AuthorizationError` file.

app/services/okta/token_validation_service.rb
```diff
@@ -0,0 +1,40 @@
...
+        http.request(req, post_data)
+      end
+
+      response.tap { |res| res['ok'] = res.is_a? Net::HTTPSuccess }
+      raise 'Failed to validate token with Okta' unless response['ok']
+
+      JSON.parse(response.body) unless response.body.nil?
+    end
+  end
+end
```
FonzMP, line 40: is it worth adding a healthcheck here to make sure we have a connection? 
`dev -> Sandbox Okta`
`staging -> Sandbox Okta`
`production -> Prod Okta`

app/api/base.rb
FonzMP: not sure I like the other approaches. It wouldn't provide the logging functionality with it. Thinking we leave as is.....   Open for discussion though.

FonzMP:
`validate_token` added to `flipper` in all environments.

FonzMP:
`validate_token` added to `flipper` in all environments.

app/api/v0/providers.rb
```diff
@@ -41,6 +45,8 @@ class Providers < V0::Base
           requires :providerName, type: String, allow_blank: false, description: 'Name of provider'
         end
         get '/release-notes' do
+          validate_token
```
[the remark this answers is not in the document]
SlakrHakr: nah, release notes are retrieved as part of the `transformations/legacy` get endpoint above. this should be fine to leave with oauth

app/api/base.rb
```diff
@@ -40,6 +44,20 @@ def protect_from_forgery
...
       end
     end
+
+    def validate_token
+      return unless Flipper.enabled? :validate_token
+
+      raise AuthorizationError if headers['Authorization'].blank?
+
+      auth_header = headers['Authorization'].split
+      raise AuthorizationError if auth_header.count < 2
```
SlakrHakr, line 54: worth checking for the word `Bearer` here?

SlakrHakr: worth checking if auth_header.count is greater than 2 ?

app/api/v0/consumers.rb
```diff
@@ -68,6 +68,8 @@ def sandbox_signup_event_content
     resource 'consumers' do
       desc 'Lists all kept consumers'
       get '/' do
+        validate_token
```
SlakrHakr, line 71: its too bad each of these have to explicitly call `validate_token`, i worry we might add an endpoint in the future and forgot to add auth, but thats probably pretty unlikely.

it is nice how explicit it is though

app/services/okta/token_validation_service.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    private
+
+    def validate_token(token)
+      auth_server_id = Figaro.env.okta_auth_server
+
+      uri = URI.parse("#{Figaro.env.okta_api_endpoint}/oauth2/#{auth_server_id}/v1/introspect")
+      req = Net::HTTP::Post.new(uri)
+
+      request(req, uri, token)
```
SlakrHakr, line 21: why not just move this up into the `token_valid?` method? is there value in this method passthru happening here that im missing?

app/services/okta/token_validation_service.rb
```diff
@@ -0,0 +1,40 @@
...
+    def token_valid?(token)
+      validate_token(token)
+    end
+
+    private
+
+    def validate_token(token)
+      auth_server_id = Figaro.env.okta_auth_server
+
+      uri = URI.parse("#{Figaro.env.okta_api_endpoint}/oauth2/#{auth_server_id}/v1/introspect")
```
SlakrHakr, line 18: wont this need to be `prod_okta_api_endpoint` in LPB production?

app/services/okta/token_validation_service.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      response = Net::HTTP.start(uri.host, uri.port, use_ssl: uri.to_s.start_with?('https')) do |http|
+        post_data = URI.encode_www_form({ token: token, token_type_hint: 'access_token' })
+        http.request(req, post_data)
+      end
+
+      response.tap { |res| res['ok'] = res.is_a? Net::HTTPSuccess }
+      raise 'Failed to validate token with Okta' unless response['ok']
+
+      JSON.parse(response.body) unless response.body.nil?
```
SlakrHakr, line 37: does the response body have valid scopes for the token? we would need each endpoint to define which scopes it requires and then validate the token has that level of access
```
    consumer.read
    consumer.write
    provider.read
    provider.write
```

app/api/v0/consumers.rb
SlakrHakr: based on https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/files#r892726884
im guessing we'd want to pass in the scope required to access each endpoint here

spec/api/v0/consumers_spec.rb
```diff
@@ -199,6 +199,28 @@
     end
   end
 
+  describe 'enforces auth for none dev portal routes' do
```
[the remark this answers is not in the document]
FonzMP: [e8613ca](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/commits/e8613ca1b16497431d36ab105d1d9718d17aa7ab)

spec/api/v0/consumers_spec.rb
```diff
@@ -199,6 +199,28 @@
...
+      VCR.use_cassette('okta/access_token_200', match_requests_on: [:method]) do
+        get '/platform-backend/v0/consumers', params: {}, headers: { Authorization: 'Bearer t0k3n' }
+        expect(response.status).to eq(200)
+      end
+    end
+
+    it 'receives unauthorized without respective token' do
+      get '/platform-backend/v0/consumers'
+      expect(response.status).to eq(401)
+    end
```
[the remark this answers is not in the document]
FonzMP: [e8613ca](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/commits/e8613ca1b16497431d36ab105d1d9718d17aa7ab)

app/services/okta/token_validation_service.rb
FonzMP: [2fdcbd7](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/commits/2fdcbd72a357e47c9585e1889be5d38deac39a28)

app/services/okta/token_validation_service.rb
FonzMP: yep. totally spaced it.

app/services/okta/token_validation_service.rb
FonzMP: valid point. It should be the same scenario cause we're using the same values. 

I'm assuming since it was `Net::HTTP` vs `Oktakit`, I was just thinking we should add a test in the healthcheck. Probably just overanalyzing on my part.

app/api/base.rb
FonzMP: [f3f9a2d](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/commits/f3f9a2d34ebd3c20356a9931b9baa8fd446964d6)

SlakrHakr: this logic still runs into the issue of someone sending in something like this:
```
Bearer 34234234 fdsdfds
```
as the auth header. this logic would get the middle section of text and use that as the token

i was more thinking using the regex kind of like this, so it always gets everything after the "Bearer " bit
```
token = 'Bearer 34234234 fdsdfds'.match(/^Bearer (.*)$/)&.captures&.first
```
and token will be nil if the passed in string doesnt match the regex

app/services/okta/token_validation_service.rb
FonzMP: [6a4489c](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/commits/6a4489cf123a35bfb2bcf0180e14e1c20b4f87ff)

app/api/v0/consumers.rb
FonzMP: [6a4489c](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/commits/6a4489cf123a35bfb2bcf0180e14e1c20b4f87ff)

app/api/base.rb
FonzMP: [6a4489c](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/commits/6a4489cf123a35bfb2bcf0180e14e1c20b4f87ff)

app/services/okta/token_validation_service.rb
FonzMP: [e848200](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/commits/e848200739655a408feb4a5040c0bf0471f51986)

app/api/v0/providers.rb
```diff
@@ -26,6 +28,8 @@ class Providers < V0::Base
                                allow_blank: true
       end
       get '/transformations/legacy' do
+        validate_token(Scope.provider_read)
```
SlakrHakr, line 31: this one we dont want to oauth

spec/api/v0/providers_spec.rb
```diff
@@ -55,4 +55,22 @@
...
+      Flipper.enable :validate_token
+    end
+
+    after do
+      Flipper.disable :validate_token
+    end
+
+    it 'retrieves consumers with access_token' do
+      VCR.use_cassette('okta/access_token_200', match_requests_on: [:method]) do
+        get '/platform-backend/v0/providers/transformations/legacy', params: {},
```
SlakrHakr, line 70: probably need to change this url to `/platform-backend/v0/providers` after removing oauth from this endpoint

app/api/v0/providers.rb
FonzMP: [770b466](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/commits/770b46671858b3e12b964135f603a7d4c58e10e9)

spec/api/v0/providers_spec.rb
FonzMP: [770b466](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/143/commits/770b46671858b3e12b964135f603a7d4c58e10e9)

SlakrHakr (approved):

PR #144, opened 2022-06-08
improve readme as new developer starts using for first time

SlakrHakr:


README.md
```diff
@@ -42,8 +52,6 @@ export RAILS_ENV=development
 rake db:create
 rake db:migrate
 rake db:seed
-rake dynamo:seed
-rake kong:seed
 rake elasticsearch:seed
```
FonzMP, line 55: why are we removing the dynamo and kong seed but not ES?

FonzMP: or was the intent to use the `seed_services` ?

guessing we may want to create a rake task that does them all if needed. Utilizing `Utility::SeedService.seed_services`

SlakrHakr: i figured we didnt need dynamo or kong seeds in this context anymore, because nothing is being pulled from them outside of the import from dynamo and the utility endpoint to pull kong consumers directly.

elasticsearch is still a thing considering we have a `statistics` endpoint that connects up with elasticsearch, which is why i left it in.

i can add them all back in though, was just trying to reduce complexity up front now that our official consumer import in production is complete

SlakrHakr: went ahead and added them back in, no sense removing them if theres a chance it might be more confusing for developers trying to use this

FonzMP: Makes sense, but with that being said, I'd argue we remove the elasticsearch seed if that were the case, because the only true way to utilize the data being created within ES is to add the creation of specific consumers (within LPB) with the necessary `gateway_ref`'s or `oauth_ref`'s. Currently we're providing data to ES that is reliant on that, so it's irrelevant without said consumer data.

FonzMP: besides, this is technically within the `Getting Started` section, and it's probably not a necessity within getting the application running.

Maybe we add another section about the seeds and what they will achieve?

FonzMP (approved):
:shipit:

FonzMP (approved):
:shipit:

PR #151, opened 2022-06-15
Add provider endpoint

FonzMP:
to make this work with grape-swagger as well as a default postman request, the two array types had to be comma-delimited and use a `coerce_with` to modify a string to an array.

### DIFF
--- a/app/api/v0/providers.rb
+++ b/app/api/v0/providers.rb
@@ -4,6 +4,14 @@ module V0
   class Providers < V0::Base
     version 'v0'
 
+    helpers do
+      def oauth_to_json
+        oauth = params['api_metadatum_attributes']['oauth_info'].to_json
+        params['api_metadatum_attributes']['oauth_info'] = oauth
+        params
+      end
+    end
+
     resource 'providers' do
       desc 'Return list of API providers', {
         headers: {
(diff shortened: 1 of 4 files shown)


app/api/v0/providers.rb
```diff
@@ -27,6 +35,69 @@ class Providers < V0::Base
...
+          optional :oauth_info, type: Hash do
+            optional :ccgInfo, type: Hash do
+              optional :baseAuthPath, type: String, allow_blank: false
+              optional :productionAud, type: String, allow_blank: true
+              optional :sandboxAud, type: String, allow_blank: true
+              optional :scopes, type: Array[String], allow_blank: false,
+                                description: 'Scopes available - comma separated',
+                                coerce_with: lambda { |v|
+                                               v.split(',')
+                                             }
```
FonzMP, line 75: we could probably find a way to nest `:scopes` and `baseAuthPath` since they're shared, but I didn't know if that was necessary. juice probably isn't worth the squeeze.

app/api/v0/providers.rb
```diff
@@ -27,6 +35,69 @@ class Providers < V0::Base
...
+            requires :id, type: Integer, values: ->(v) { ApiCategory.kept.map(&:id).include?(v) }
+          end
+        end
+      end
+      post '/' do
+        validate_token(Scope.provider_write)
+
+        api = Api.create(name: params.delete(:name))
+        oauth_to_json if params[:api_metadatum_attributes][:oauth_info].present?
+        api.update(params)
```
FonzMP, line 96: I wanted this all within the create but it didn't want to generate anything besides the API when done that way, so I mocked the `seeds.rb` doing a `.create` and `.update`. Noticed in the UI for the DB that the `oauth_info` looked a bit different so modified it to be converted to json on our end. Looks like everything is being generated from the same structure as an object from the `seeds.rb` file, so consistency is king? 👑

SlakrHakr (approved):

PR #157, opened 2022-06-22
[API-16913] - Update va_internal_only to be more than a boolean

gileswells:
https://vajira.max.gov/browse/API-16913

Originally planned as a true enum the end result was an enum in the data model but a string field in the DB due to unit testing oddities with type casting / magic rails shenanigans causing issues.


SlakrHakr (approved):
LGTM

PR #161, opened 2022-06-23
remove fields from entity

FonzMP:
removes fields that are no longer included within the api_env


spec/api/utilities_spec.rb
```diff
@@ -5,7 +5,7 @@
 describe Utilities, type: :request do
   describe 'APIs' do
     before do
-      create(:api, name: 'Claims API')
+      create(:api, :with_api_environment)
```
FonzMP, line 8: name was irrelevant in test

SlakrHakr (approved):

PR #164, opened 2022-06-30
Api 17185 utilities auth server

FonzMP:
This will require a manual creation within `/admin/database` for each environment for an API and an ApiRef. I added LPB to the seeds file for local though. The respective ApiRef will need the `name` of `lpb`. 

Brakeman was failing so I ran a `bundle update` and this is what we got. Some of the rubocop rules had to be adjusted (autocorrect and DeprecatedAttributeAssignment) due to that. This update also impacted some spacing prior to `\`. These were all auto corrected by rubocop.


app/api/utilities.rb
```diff
@@ -3,6 +3,22 @@
...
+  helpers do
+    def user_from_signup_params
+      user = User.find_or_initialize_by(email: params[:email])
+
+      user.first_name = params[:firstName]
+      user.last_name = params[:lastName]
+      user.consumer = Consumer.new if user.consumer.blank?
+      user.consumer.description = params[:description]
+      user.consumer.organization = params[:organization]
+      user.consumer.apis_list = ApiService.parse('ccg/lpb', filter_lpb: false)
```
FonzMP, line 15: force setting this to only work with LPB for this particular endpoint. We could pass in `ccg/lpb` via an apis param if we wanted, but I thought this would make it easier currently.

app/api/utilities.rb
```diff
@@ -66,6 +82,30 @@ class Utilities < Base
         drift_service_arg = params[:environment] == 'production' ? :production : nil
         Okta::DriftService.new(drift_service_arg).detect_drift(alert: false, filter: params[:filterLastDay])
       end
+
+      namespace 'lpb' do
+        params do
+          requires :environment, type: Symbol, values: [:production], allow_blank: false
```
FonzMP, line 88: I wanted to originally set this to include `sandbox` and `production` but we would have to rework how the `auth_server_access_key` works with the APIs currently. Due to this providing signups essentially to allow access, and providing an individual the ability to retrieve an access_token, I guess it makes sense to only use `production` currently.

app/api/utilities.rb
```diff
@@ -66,6 +82,30 @@ class Utilities < Base
...
+          optional :oAuthPublicKey, type: JSON
+          requires :organization, type: String
+          requires :termsOfService, type: Boolean, allow_blank: false, values: [true]
+        end
+        post 'applications' do
+          user = user_from_signup_params
+
+          okta_service = Okta::ServiceFactory.service(params[:environment])
+          okta_consumers = okta_service.consumer_signup(user,
+                                                        { application_public_key: params[:oAuthPublicKey] })
```
FonzMP, line 102: originally added the `okta_signup_options` method to the `utilities.rb` as well, but it didn't make sense to have a 5 line method for one used item. This is a little ugly in respect to formatting, but rubocop didn't like a one liner so I split the service and the actual signup to accommodate. I'm not set on this, glad to do a one liner and add a rubocop exclusion, or move back to the `okta_signup_options` method.

app/api/utilities.rb
```diff
@@ -66,6 +82,30 @@ class Utilities < Base
...
+          Event.create(event_type: Event::EVENT_TYPES[:lpb_signup], content: { user: user, okta: okta_consumers })
```
FonzMP, line 103: just added user and okta details so we'll have them stored in the events table. We could join these but thought it made sense to keep the details separate. Okta is a little bulky, glad to trim some of that down as well if needed.

app/api/v0/consumers.rb
```diff
@@ -116,7 +116,7 @@ def subscribed?(consumers)
       params do
         requires :apis, type: Array[Api],
                         allow_blank: false,
-                        coerce_with: ->(value) { ApiService.parse(value) },
+                        coerce_with: ->(value) { ApiService.parse(value, filter_lpb: true) },
```
FonzMP, line 119: rubocop wanted me to set a default boolean to use `filter_lpb: true` vs `filter_lpb = true` which forces me to pass the entirety of the value (so `filter_lpb: true` vs `true` or for that matter, any value if it were the default). I don't like that. But with only 3 places it's being used, it wasn't overwhelming so? With this, I think it makes sense to not use a default and apply true or false? granted this provides a little clarity on what it's doing.

app/services/api_service.rb
```diff
@@ -3,14 +3,14 @@
...
-  def self.parse(api_list_str)
+  def self.parse(api_list_str, filter_lpb: true)
     api_list = api_list_str.strip.split(',')
     api_list.map do |api|
       api_parts = api.strip.split('/')
       raise invalid_api_exception(api) if api_parts.blank? || api_parts.count > 2
 
       api_ref = ApiRef.kept.find_by(name: api_parts.last)
-      raise invalid_api_exception(api) if api_ref.blank?
+      raise invalid_api_exception(api) if api_ref.blank? || (api_ref.name.eql?('lpb') && filter_lpb)
```
FonzMP, line 13: I could add this stuff to a method and pass the api to make it a little cleaner, but probably just as bulky to add another method specific to this?

app/api/utilities.rb
```diff
@@ -66,6 +82,30 @@ class Utilities < Base
         drift_service_arg = params[:environment] == 'production' ? :production : nil
         Okta::DriftService.new(drift_service_arg).detect_drift(alert: false, filter: params[:filterLastDay])
       end
+
+      namespace 'lpb' do
```
FonzMP, line 86: open to changes on this as it comes out to `/platform-backend/utilities/okta/lpb/applications`. I just didn't really like duplicating `platform-backend` and wanted something to notate that it's LPB specific. Glad to modify to whatever we feel appropriate.

app/api/utilities.rb
SlakrHakr: i think having okta_signup_options in this class makes sense overall. yeah, its just a single line at the moment but it 
1. provides consistency across the application and
2. allows for growth if we need to send more in the future

that being said, this isnt the end of the world either, can always iterate and move this out to a separate method in the future if need be

app/api/v0/consumers.rb
SlakrHakr: 1. i think this approach makes it clearer than just passing in a boolean, so this makes sense to me
2. i think a default of true actually makes sense to me here, it should be the default and this one utility endpoint should be the only exception

app/services/api_service.rb
SlakrHakr: this seems fine to me 👍🏻

SlakrHakr: not super important here, but worth mentioning

if `api_ref.name.eql?('lpb') && filter_lpb` was switched to be `filter_lpb && api_ref.name.eql?('lpb')` the vast majority of the time the string equality bit wouldnt need to execute.

SlakrHakr (dismissed):

SlakrHakr (approved):

PR #165, opened 2022-07-07
better error message if invalid jwk is provided

SlakrHakr:
before this change, if the provided oauth public key for ccg was json but ultimately invalid we would respond with an error from okta.

this change introduces our own basic validation so we can provide a better error message to a consumer on the frontend.


app/api/validators/provided_oauth_params.rb
```diff
@@ -25,10 +25,23 @@ def validate_acg_oauth_params!(params)
...
-                                              message: 'missing required oAuth value')
+      if params[:oAuthPublicKey].blank?
+        raise Grape::Exceptions::Validation.new(params: %w[oAuthPublicKey],
+                                                message: 'missing required oAuth value')
+      end
+
+      begin
+        jwk = JSON.parse(params[:oAuthPublicKey])
+      rescue JSON::ParserError
+        return # this will get caught and responded to by grape's built-in validator
```
SlakrHakr, line 36: initially was raising a validation exception here as well but it resulted in this error plus the generic one provided by grape in the response. this made it so there is always only one error message in the response for this value

app/api/validators/provided_oauth_params.rb
```diff
@@ -25,10 +25,23 @@ def validate_acg_oauth_params!(params)
...
+      end
+
+      begin
+        JWT::JWK.import(jwk)
```
SlakrHakr, line 40: i wanted to avoid us pre-validating the jwk here in case our validation caused a valid submission to fail for whatever reason. made more sense to have okta tell us whether the jwk was allowed or not.

unfortunately, looking through the error message we get from okta, the only way to try to detect this would be to check against the message okta returns, which seems brittle and worse than this approach imo.

this validation seems simple enough to hopefully not bite us later on

FonzMP: yeah, I'm a little scared of this to be honest. Do we not have an error code that will stay consistent?

PR #166, opened 2022-07-13
Api 17401 lpb token validation

FonzMP:


app/services/okta/token_validation_service.rb
```diff
@@ -18,17 +12,15 @@ def token_valid?(token)
     private
 
     def request(req, uri, token)
-      client_id = Figaro.env.okta_client_id
-      client_secret = Figaro.env.okta_client_secret
-      req.basic_auth(client_id, client_secret)
+      req['Authorization'] = "Bearer #{token}"
+      req['apikey'] = Figaro.env.validate_token_apikey
+      req.set_form_data({ aud: Figaro.env.okta_audience })
```
FonzMP, line 17: these are needed by the `validate_token` service.

FonzMP:
this PR includes a `bundle update` for brakeman and `rubocop` updates to align with that.

FonzMP:
accidentally closed 👍

app/services/okta/token_validation_service.rb
```diff
@@ -2,14 +2,8 @@
...
   class TokenValidationService
-    KEYS = {
-      alg: 'RS256'
-    }.freeze
-
     def token_valid?(token)
-      auth_server_id = Figaro.env.okta_auth_server
-
-      uri = URI.parse("#{Figaro.env.okta_token_validation_endpoint}/oauth2/#{auth_server_id}/v1/introspect")
+      uri = URI.parse("#{Figaro.env.prod_kong_token_validation}/internal/auth/v2/validation")
```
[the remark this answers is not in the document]
SlakrHakr: it means we'll have to add a bunch more duplicated environment variables in aws then which is unfortunate...
also is unfortunate to have to make an implementation change to fix a testing problem

i wonder if it makes "more" sense to try to "not ignore?" this specific path within vcr somehow? "ignore all localhost except this specific path" type thing? at least then it would be handled within the testing framework entirely.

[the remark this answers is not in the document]
FonzMP: I could take a second to peruse the documentation and maybe we can bypass the `ignore_hosts` for the 3 tests associated to this? We can't really do it across all `prod_kong_elb` cause we'll end up causing issues with other tests. I'll take a quick gander.

FonzMP:
Updated README with some basic test information.

spec/rails_helper.rb
```diff
@@ -30,9 +30,9 @@ def self.all_matches
   c.hook_into :webmock
   c.default_cassette_options = { decode_compressed_response: true }
   c.ignore_request do |request|
-    URI(request.uri).port.in?([8000, 8001, 9200])
+    URI(request.uri).port.in?([8000, 8001, 9200]) && !URI(request.uri).path.in?(['/internal/auth/v2/validation'])
   end
-  c.ignore_hosts '127.0.0.1', 'localhost'
+  c.ignore_localhost = true
```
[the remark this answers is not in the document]
FonzMP: nvm [683d7f8](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/166/commits/683d7f8f26732941f96a646889b362387c138f16)

app/api/base.rb
```diff
@@ -55,8 +55,8 @@ def validate_token(scope)
 
       response = Okta::TokenValidationService.new.token_valid?(token)
 
-      raise AuthorizationError unless response['active']
-      raise ForbiddenError unless response['scope'].include?(scope)
+      raise AuthorizationError if response['errors'].present?
+      raise ForbiddenError unless response['data']['attributes']['scp'].include?(scope)
```
SlakrHakr, line 59: i was going to suggest something less prone to null errors like
```
      raise ForbiddenError unless response.dig('data', 'attributes', 'scp')&.include?(scope)
```

the more i thought about it though, we probably want it to be a 500 error if the response here is formed different than we expect.

i wonder if it would still be worth doing something like this:
```
response_scope = response.dig('data', 'attributes', 'scp')
raise 'auth scope not found' if response_scope.blank?
raise ForbiddenError unless response_scope.include?(scope)
```

that would still give the user a 500 (because its something wrong with the system, not something they can fix in their request) but it would give us a much more descriptive error message

FonzMP: `.dig` does make this approach a bit better. I didn't realize it would automatically return `nil` if any intermediate step is missed. 

I can add these changes right quick.

app/services/okta/token_validation_service.rb
```diff
@@ -18,17 +13,15 @@ def token_valid?(token)
...
+      req.set_form_data({ aud: Figaro.env.okta_audience })
 
       response = Net::HTTP.start(uri.host, uri.port, use_ssl: uri.to_s.start_with?('https')) do |http|
-        post_data = URI.encode_www_form({ token: token, token_type_hint: 'access_token' })
-        http.request(req, post_data)
+        http.request(req)
       end
 
       response.tap { |res| res['ok'] = res.is_a? Net::HTTPSuccess }
-      Rails.logger.warn("??? #{response}") && raise('Failed to validate token with Okta') unless response['ok']
```
SlakrHakr, line 31: what happens if calling the validation endpoint doesnt return a 200 status? i dont see anything else that would capture that issue and raise an exception, but i definitely could be missing something

SlakrHakr (dismissed):
everything looks good to me except this question:
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/166/files#r922428431

and it sounds like we might need an additional environment variable for "kong validation endpoint" due to deployed environments having additional path associated with the existing kong environment variables

FonzMP:
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend-deployment/pull/27

FonzMP:
> everything looks good to me except this question: https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/166/files#r922428431
> 
> and it sounds like we might need an additional environment variable for "kong validation endpoint" due to deployed environments having additional path associated with the existing kong environment variables

https://github.com/department-of-veterans-affairs/lighthouse-platform-backend-deployment/pull/27

app/api/base.rb
```diff
@@ -55,8 +55,11 @@ def validate_token(scope)
 
       response = Okta::TokenValidationService.new.token_valid?(token)
```
SlakrHakr, line 56: might be a bit nitpicky, but all this discussion around this `token_valid?` method and at one point i was confused due to the ruby standard for a method that ends in `?` to only return either true or false.

options would be to either:
1. move the logic below into that method (problem with that is a token can either be unauthorized or forbidden)
2. rename the method to something like `validate_token` without the question mark, leave all the below logic here in this class
3. rename method to `validate_token!` and move all the below logic into it and have it either raise an exception or return null on success

FonzMP: [b6a7233](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/166/commits/b6a72337022738555a4c075f2094c4060d2a31a4)

app/services/okta/token_validation_service.rb
FonzMP: [1fc5918](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/166/commits/1fc591822451bd268736347c5530768f8fb7a709)

SlakrHakr (approved):

PR #169, opened 2022-07-18
API-17493: Investigate and Fix Slack Sandbox Signup Reporting Issue

SlakrHakr:
new event types are causing issues with this query
all sql queries need to filter down to this specific event type


FonzMP (approved):
:shipit:

PR #170, opened 2022-07-18
Filter report values

FonzMP:
adds va_auth to the seeds and a temp filter to eliminate the next report from displaying LPB and VA Auth APIs.

We'll need a way to remove these from the overall query, but this will stop the query per API.


FonzMP:
the `event_type` filter will remove LPB as they're `lpb_signups`, but va_auth goes through the `sandbox_signup` currently. we could move this to the lpb route (or it's own) and just make the `event_type` dependent on the API. Which would remove the `sandbox_signup` type and place it in it's own category. Or swap both of them to something like `internal_signup`.

app/services/slack/report_service.rb
```diff
@@ -15,6 +15,10 @@ def query_events(span, start_date)
 
     private
 
+    def excluded_api_list
+      %w[lpb va_auth].freeze
```
SlakrHakr, line 19: one method we could use to get around defining specific apis here is just checking if the api has apimetadatum associated with it.

theoretically, i would assume we only want apis in our reports that are present on the developer-portal, and every api present on developer.va.gov would have an api-metadatum record tied to it

FonzMP: [d3d7602](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/170/commits/d3d760228d9f5b83f5d3243e796ffe5ab5f5d918)

app/services/slack/report_service.rb
```diff
@@ -23,7 +23,7 @@ def ref_builder(ref)
     end
 
     def sort_refs
-      ApiRef.all.map(&:name).uniq.sort
+      ApiRef.select { |ref| ref unless ref.api.api_metadatum.nil? }.map(&:name).uniq.sort
```
SlakrHakr, line 26: nit:

```
ApiRef.select { |ref| ref.api.api_metadatum.present? }
```
reads more clearly to me personally

that being said, i tested this locally and it works the same way. so functionality definitely works as expected

SlakrHakr (approved):

PR #175, opened 2022-08-01
update okta error handling

FonzMP:
Adding some error handling from the Okta responses to hopefully provide better specification in to what the issue may be.


app/services/okta/base_service.rb
```diff
@@ -128,6 +130,14 @@ def build_new_application_payload(user, type, options = {})
       raise 'Invalid supplied arguments'
     end
 
+    def handle_okta_error(application)
+      app = application.to_h
+      summary = app[:errorSummary]
+      cause = app[:errorCauses]&.first&.dig(:errorSummary)
+      response_cause = " - #{cause}" unless cause.nil?
```
SlakrHakr, line 137: use .blank? instead of .nil? in case cause is ever an empty string for example

SlakrHakr (approved):

app/services/okta/base_service.rb
FonzMP: [4bb28ca](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/175/commits/4bb28ca6ee413ea824041a143b11f5b36e869788)

SlakrHakr (approved):

PR #178, opened 2022-08-02
API-17549: enable utilities endpoints on-demand

SlakrHakr:
Working with backstage, it provides swagger-ui for free.
The problem is the utilities endpoints are shown there and can be interacted with without being locked down to our team in production.
This is the best solution I could come up with. It means we will have to enable the utilities endpoints when we need to use them and disable them (also hides them from the openapi spec) when theyre not in use.

I tried to tie showing or hiding the endpoints to the signed in user, but I couldn't get access to the current_user (or the request at all) at the stage where hiding needed to be determined.

### DIFF
--- a/app/services/okta/base_service.rb
+++ b/app/services/okta/base_service.rb
@@ -134,7 +134,7 @@ module Okta
       app = application.to_h
       summary = app[:errorSummary]
       cause = app[:errorCauses]&.first&.dig(:errorSummary)
-      response_cause = " - #{cause}" unless cause.blank?
+      response_cause = " - #{cause}" if cause.present?
       summary + (response_cause || '')
     end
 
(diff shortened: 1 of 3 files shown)


app/services/okta/base_service.rb
```diff
@@ -134,7 +134,7 @@ def handle_okta_error(application)
       app = application.to_h
       summary = app[:errorSummary]
       cause = app[:errorCauses]&.first&.dig(:errorSummary)
-      response_cause = " - #{cause}" unless cause.blank?
+      response_cause = " - #{cause}" if cause.present?
```
SlakrHakr, line 137: rubocop complained, im surprised it got into master actually 🤷🏻

PR #182, opened 2022-08-08
Api 17685 apigee import

FonzMP:
Builds the general list pulling directly from Kong and Okta and building from that data, as LPB does not have the entirety of consumers due to the ability to create consumers outside of LPB (Kong CLI, Okta Interface). 

This will, unfortunately, retrieve all consumers across all environments in respect to Okta for each LPB environment, but we do not really have a way to discern consumers from various environments. 

There are updates to the [oktakit branch](https://github.com/FonzMP/oktakit/tree/auth_servers_and_secrets) we're using that are required as we needed a way to retrieve secrets using Oktakit vs Net::HTTP.

Currently the layout is, pull all Kong consumers from `/consumers`, pull all acls from `/acls` and all keys from `key-auths` and use these as a filterable list, therefore only requiring three calls to Kong. The Okta portion is a bit more complex. We currently have to retrieve all `applications`, then all `auth servers` AND `auth server policies`. Once we return that, if the client app is listed as `client_secret_basic`, we need to make ANOTHER request to `v1/applications/#{app[:id]}/credentials/secrets` PER application. This is the concern as it will be rather slow AND we'll have to consider rate limiting from Okta.


FonzMP:
there have been requested changes to the structure of the data. I'll be working on making those changes.

jw81:
With the recent announcement that we will no longer be integrating with Apigee, it looks like we won't be needing this logic anymore. I believe you can now close this PR @FonzMP

FonzMP:
👋 Apigee

PR #185, opened 2022-08-15
textarea fields in rails admin view

SlakrHakr:
tested locally with existing content and all content was fine after the migration

available rails_admin types: https://github.com/rs-pro/rails_admin_settings#value-types

### DIFF
--- a/db/migrate/20220815213849_convert_fields_with_new_lines.rb
+++ b/db/migrate/20220815213849_convert_fields_with_new_lines.rb
@@ -1,0 +1,13 @@
+class ConvertFieldsWithNewLines < ActiveRecord::Migration[7.0]
+  def up
+    change_column :api_categories, :overview, :text
+    change_column :api_categories, :quickstart, :text
+    change_column :api_release_notes, :content, :text
+  end
+
+  def down
+    change_column :api_categories, :quickstart, :string
+    change_column :api_categories, :quickstart, :text
+    change_column :api_release_notes, :content, :text
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20220815213849_convert_fields_with_new_lines.rb
```diff
@@ -0,0 +1,13 @@
+class ConvertFieldsWithNewLines < ActiveRecord::Migration[7.0]
+  def up
+    change_column :api_categories, :overview, :text
+    change_column :api_categories, :quickstart, :text
+    change_column :api_release_notes, :content, :text
+  end
+
+  def down
+    change_column :api_categories, :quickstart, :string
```
FonzMP, line 9: is this supposed to be `:overview` ?

db/migrate/20220815213849_convert_fields_with_new_lines.rb
```diff
@@ -0,0 +1,13 @@
...
+    change_column :api_categories, :quickstart, :text
+    change_column :api_release_notes, :content, :text
```
FonzMP, line 11: wouldn't we want the rollback to be `:string` or do you want to force them to remain `:text` by adding this?

SlakrHakr: yes :(
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/185/commits/998d7c312dcfd6c2c439cc58737be29981606949

FonzMP (approved):
:shipit:

PR #188, opened 2022-08-16
convert auth type to camelcase

SlakrHakr:


FonzMP:
there's a few other spots. For example
https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/45bc12d7d614c555d35191e5fa1fd8c25baaaae0/app/api/v0/providers.rb#L34

unless we want to format it when it comes in. something like grapes example
`requires :email_address, as: :email`

so something like
`optional :authType, as: :auth_type, : type: String, values: %w[apikey oauth oauth/acg oauth/ccg]` ?

FonzMP:
yucky.. but then we need something like this if we want to use the `as: :auth_type`
```
declared = declared(params)
if declared[:auth_type].present?
```

probably worth just changing all of them to `:authType` ?!?!?!

FonzMP (approved):
:shipit:

PR #208, opened 2022-09-08
Api 14772 auto deploy

FonzMP:


.github/workflows/assets/maintenance-request.md
```diff
@@ -0,0 +1,49 @@
+---
+name: Maintenance Request
+about: Request a change to Lab/Sandbox or the Production environment.
+title: __mr_summary__ (YYYY/MM/DD HH:MM ET)
+labels: maintenance-request
+assignees: FonzMP
```
FonzMP, line 6: this will change to gabe to match dev-portal once we're ready. for now it is using my github for testing.

.github/workflows/auto-deploy-to-production.yml
```diff
@@ -0,0 +1,156 @@
+name: Auto Deploy master to Production
+
+on:
+  # workflow_dispatch:
+  # schedule:
+  #   - cron: '*/30 13,14,15,16,17,18,19,20,21,22,23 * * 1,2,3,4,5'
```
FonzMP, line 6: this will need to be enabled when we're ready to do a new MR and test against LPB issue.

.github/workflows/auto-deploy-to-production.yml
```diff
@@ -0,0 +1,156 @@
...
+  push:
+
+env:
+  environment: 'QA'
```
FonzMP, line 10: for testing. will move to production if it works correctly.

.github/workflows/auto-deploy-to-production.yml
```diff
@@ -0,0 +1,156 @@
...
+          fi
+
+          echo 'Successfully queued for build...'
+          echo "::set-output name=build_number::$NEXT_BUILD"
+      - id: check_build_status
+        name: Check Build Status
+        if: steps.deploy_to_production.outcome == 'success'
+        run: |
+          echo "Allowing queued jobs to be picked up... temporarily sleeping..."
+          sleep 60
```
FonzMP, line 87: there can be three jobs in the queue and jenkins has a 5 minute time frame to pick them up. I wonder if we move to 300 seconds just to be safe? that being said, we may not get the next build number, that's why we'll increment below until we get a 404 or a non 200 response. Hopefully this is a safe way to break the `while` if we don't locate the build in the builds array.

.github/workflows/auto-deploy-to-production.yml
```diff
@@ -0,0 +1,156 @@
...
+          while [ "$BUILD_FOUND" == "false" ]; do
+            STATUS=`curl  -w "%{http_code}" -o build.json -u ${{ secrets.user_jenkins }}:${{ secrets.api_token_jenkins }} https://tools.health.dev-developer.va.gov/jenkins/job/department-of-veterans-affairs/job/health-apis-deployer/job/d2/${{ build_number }}/api/json`
+            if [ $STATUS -ne 200 ]; then
+              echo 'Failed to locate build...'
+              echo 'exiting...'
+              exit 1;
+            fi
+            USER_ID=`jq '.actions[] | select(._class == "hudson.model.CauseAction") | .causes[0].userId' build.json`
+            if [ "$USER_ID" == "okapibot" ]; then
+              BUILD_FOUND="true"
```
FonzMP, line 98: this seemed like the safest / easiest way. we could match against the VPC and the production as well, if we feel it necessary, but only production builds to LPB should be using okapibot, so it SHOULD signify the correct build.

.github/workflows/auto-deploy-to-production.yml
```diff
@@ -0,0 +1,156 @@
...
+            else
+              ((build_number++))
+            fi
+          done
+
+          RUNNING="true"
+          while [ $RUNNING == "true" ]; do
+            STATUS=`jq '.result' build.json`
+            if [ "$STATUS" == "FAILURE" ]
+              echo 'Oh noes, something borked...'
```
FonzMP, line 108: had to.

.github/workflows/auto-deploy-to-production.yml
```diff
@@ -0,0 +1,156 @@
...
+              echo "https://tools.health.dev-developer.va.gov/jenkins/job/department-of-veterans-affairs/job/health-apis-deployer/job/d2/$build_number"
+              exit 1;
+            fi
+            if [ "$STATUS" == "SUCCESS" ]
+              RUNNING="false"
+            else
+              sleep 60
+              curl -u ${{ secrets.user_jenkins }}:${{ secrets.api_token_jenkins }} https://tools.health.dev-developer.va.gov/jenkins/job/department-of-veterans-affairs/job/health-apis-deployer/job/d2/${{ build_number }}/api/json > build.json
```
FonzMP, line 116: continuously fetches the status to update the MR once we have a failure or success. I haven't caught the fields value while building, assuming it's not `SUCCESS` or `FAILURE`

.github/workflows/create-auto-deploy-maintenance-request.yml
```diff
@@ -0,0 +1,113 @@
+name: Create Auto Deploy Maintenance Request
+
+on:
+  push:
```
FonzMP, line 4: set to push, and check for the 'Snapshot' commit message. initial thought was to use the 'Next Snapshot' commit sent by libertybot, but to truly treat it as an auto deploy on a new release, we should probably aim for `on.release`. That being said, that can only be ran on `master` so it won't work for testing.

.github/workflows/auto-deploy-to-production.yml
SlakrHakr: feels like there is a risk in this, but low risk probably
should be an easy thing to discover and resolve if there is an issue in the future though

PR #221, opened 2022-09-16
Api 18680/signups to apigee

FonzMP:
Probably needs some cleanup but working state of Apigee signup flow.


app/api/v0/consumers.rb
```diff
@@ -146,10 +146,16 @@ def subscribed?(consumers)
 
         user = user_from_signup_params
 
+        # check user before saving for apigee route select
+        apigee_route = user.persisted? ? 'keys' : 'consumer'
```
FonzMP, line 150: route will be determined by user existing in the DB prior to signup. Check here before user is saved when attaching api_assignments

app/api/v0/consumers.rb
```diff
@@ -146,10 +146,16 @@ def subscribed?(consumers)
...
+
         kong_consumer = Kong::ServiceFactory.service(:sandbox).consumer_signup(user)
         okta_consumers = Okta::ServiceFactory.service(:sandbox).consumer_signup(user, okta_signup_options)
         Event.create(event_type: Event::EVENT_TYPES[:sandbox_signup], content: sandbox_signup_event_content)
 
+        _apigee_consumer = Apigee::ServiceFactory.service(:sandbox).consumer_signup(user, kong_consumer,
```
FonzMP, line 156: not sure if we even care about the variable since it's unused, but I'm sure we may incorporate it later down the line. Currently it returns the same data we're submitting, so not much we can do with it (no ID's or anything associated with the Apigee success)

app/models/api.rb
```diff
@@ -36,6 +36,10 @@ def locate_auth_types
     types.uniq
   end
 
+  def internal_only?
+    api_metadatum.present? && api_metadatum.va_internal_only == 'StrictlyInternal'
+  end
```
FonzMP, line 41: assuming these Apis are the ones that Apigee will host.

app/services/apigee/base_service.rb
```diff
@@ -0,0 +1,119 @@
...
+      request
+    end
+
+    def append_consumer_data(request, user, kong_consumer, okta_consumers)
+      request[:developer][:firstName] = user.first_name
+      request[:developer][:lastName] = user.last_name
+      request[:developer][:userName] = define_username(kong_consumer, okta_consumers)
+
+      request
+    end
```
FonzMP, line 55: only needed for `/consumers`

app/services/apigee/base_service.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    def define_username(kong_consumer, okta_consumers)
+      kong_consumer&.[](:kong_username) || okta_consumers&.[](:acg)&.[](:label) || okta_consumers&.[](:ccg)&.[](:label)
```
FonzMP, line 58: interesting way of doing it. probably something cleaner here.

app/services/apigee/base_service.rb
```diff
@@ -0,0 +1,119 @@
...
+    def api_product_list(apis, type)
+      [].tap do |a|
+        apis.each { |api| a << api.name if api.auth_type == type }
+      end
+    end
+
+    def request(req, uri)
+      req['apikey'] = apigee_gateway_apikey
+      response = @client.start(uri.host, uri.port, use_ssl: uri.to_s.start_with?('https'),
+                                                   verify_mode: OpenSSL::SSL::VERIFY_NONE) do |http|
```
FonzMP, line 109: verify_mode for now to bypass. Set brakeman to ignore.

jw81:
With the recent announcement that we will no longer be integrating with Apigee, it looks like we won't be needing this logic anymore.  I believe you can now close this PR @FonzMP

FonzMP:
Goodbye Apigee. 😢

PR #222, opened 2022-09-22
[API-18379] - Adds DB tables for storing 'production request' data

jw81:
- Adds the database tables for storing data sent to the `/platform-backend/v0/consumers/production-requests` endpoint.


db/migrate/20220922185644_create_production_request_contacts.rb
```diff
@@ -0,0 +1,13 @@
+class CreateProductionRequestContacts < ActiveRecord::Migration[7.0]
+  def change
+    create_table :production_request_contacts, id: :uuid do |t|
+      t.string :first_name
+      t.string :last_name
+      t.string :email
```
gileswells, line 6: I'm not sure that I completely agree with the direction here. I get that it's contacts on a prod access request BUT in the future, if we're able to get user accounts to a place where we'd want to have a permissions based setup, then this table should be more of a reference table like `consumers` but in this case would simply have, in essence, the `production_request` and `user` references and some timestamps.

In the business logic this would be similar to a new sandbox signup where it'd create the user record if one doesn't match on the email address.

db/migrate/20220922183948_create_production_requests.rb
```diff
@@ -0,0 +1,42 @@
+class CreateProductionRequests < ActiveRecord::Migration[7.0]
+  def change
+    create_table :production_requests, id: :uuid do |t|
+      t.string :apis
```
gileswells, line 4: I can see the value of capturing the raw data but also do think we could build out the references through a many to many table to allow for easier data reporting and other metrics to be pulled.

Architecturally thinking, would something like this make more sense to store the raw from data in something like DynamoDB? I do like structure but if we do just drop a JSON of the request to Dynamo, or even a JSON field, we wouldn't necessarily need to update the LPB schema with any new fields added to the front end unless they have new relationships that would make sense to connect together on submission.

db/migrate/20220922185644_create_production_request_contacts.rb
gileswells: Keeping the raw values submitted would make sense for posterity sake if the `first_name / last_name / email` don't all exactly match up on the request vs what's in the `users` table already.

gileswells (changes requested):
Couple thoughts on structure of things with regards to potentially leveraging relationships and making front end changes potentially easier to implement.

db/migrate/20220922183948_create_production_requests.rb
jw81: Switched it over to an `apis_production_requests` many-to-many table.

jw81:
These are the migration generators I used to create these migrations.
```
rails generate migration CreateProductionRequests \
app_description:text \
app_name:string \
breach_management_process:text \
business_model:text \
centralized_backend_log:string \
distributing_api_keys_to_customers:boolean \
expose_veteran_information_to_third_parties:boolean \
is_508_compliant:boolean \
listed_on_my_health_application:boolean \
monitization_explanation:text \
monitized_veteran_information:boolean \
multiple_req_safeguards:text \
naming_convention:string \
organization:string \
phone_number:string \
pii_storage_method:text \
platforms:string \
privacy_policy_url:string \
production_key_credential_storage:text \
production_or_oauth_key_credential_storage:text \
scopes_access_requested:text \
sign_up_link_url:string \
status_update_emails:string \
store_pii_or_phi:boolean \
support_link_url:string \
terms_of_service_url:string \
third_party_info_description:text \
value_provided:text \
vasi_system_name:string \
veteran_facing:boolean \
veteran_facing_description:text \
vulnerability_management:text \
website:string \
--primary-key-type=uuid
```

```
rails generate migration CreateApisProductionRequests \
api:references \
production_request:references \
--primary-key-type=uuid
```

```
rails generate migration CreateProductionRequestsUsers \
production_request:references \
user:references \
type:string \
--primary-key-type=uuid

```

jw81:
Trying to keep the schema changes and the biz logic changes in separate PRs is becoming too much of a headache.
I've consolidated all of the work into [this PR](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/224) instead.
Closing this PR now.

PR #229, opened 2022-10-28
[API-20044] - Add veteran redirect fields to APIs

andresriveratoro:
Closes https://vajira.max.gov/browse/API-20044
This PR adds veteran redirect fields to APIs.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -273,6 +273,7 @@ benefits_reference_api.update(
    api_category_attributes: {
      id: benefits_category.id
    }
+  }
 )
 
 facilities_api = Api.create(name: 'va-facilities')
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -272,11 +272,7 @@ benefits_reference_api.update(
    url_fragment: 'benefits_reference_data',
    api_category_attributes: {
      id: benefits_category.id
-   },
-   veteran_redirect_link_text: 'Hi',
-   veteran_redirect_link_url: "https://www.google.com/",
-   veteran_redirect_message: "Hello!"
-  }
+   }
 )
 
 facilities_api = Api.create(name: 'va-facilities')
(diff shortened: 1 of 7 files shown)


db/seeds.rb
```diff
@@ -272,7 +272,10 @@
    url_fragment: 'benefits_reference_data',
    api_category_attributes: {
      id: benefits_category.id
-   }
+   },
+   veteran_redirect_link_text: 'Hi',
+   veteran_redirect_link_url: "https://www.google.com/",
+   veteran_redirect_message: "Hello!"
```
andresriveratoro, line 278: This will be removed before merging.

jw81:
Screen Shot 2022-11-01 at 3 58 12 PM

@gileswells can you confirm my understanding of what this ticket is aiming to do?

Above is a screenshot of a portion of what the [legacy.json endpoint](https://developer.va.gov/platform-backend/v0/providers/transformations/legacy.json) returns today.  It shows that there's a `benefits.apis[0].veteranRedirect` attribute (this is Benefits Claims API) and there's also a `benefits.content.veteranRedirect` attribute.  Both attributes have the same value because they're both ultimately defined by the same `ApiCategory` object.  

So is this ticket's goal to be able to set a more specific `veteranRedirect` attribute for each individual API, rather than just inheriting the `veteranRedirect` value it gets from its `ApiCategory` object?

In other words, allow for `legacy.json` endpoint to return something like this,
```
{
  "benefits": {
    "apis": [
      {
        "altId": "claims",
        // bunch of other attributes
        "veteranRedirect": {
          "linkUrl": "my super specific url",
          "linkText": "my super specific test",
          "message": "my super specific message"
        }
      }
    ],
    "content": {
      "veteranRedirect": {
          "linkUrl": "something generic from the ApiCategory object",
          "linkText": "something generic from the ApiCategory object",
          "message": "something generic from the ApiCategory object"
        }
    }
  }
}
```

gileswells:
@jw81 thats precisely the goal.

gileswells (approved):
LGTM

gileswells (approved):
LGTM

PR #230, opened 2022-11-02
[API-18364] - upload optional logo production access

gileswells:
https://vajira.max.gov/browse/API-18364

Provides a new endpoint to pre-sign uploads to resizable logos s3 bucket.


app/api/v0/consumers.rb
```diff
@@ -285,6 +288,23 @@ def subscribed?(consumers)
                       kong_consumer: kong_consumer,
                       okta_consumers: okta_consumers
       end
+
+      params do
+        requires :fileName, type: String, allow_blank: false
+        requires :fileType,
+                 regexp: { value: %r{^image/(jpeg|png)$},
```
imcgonigle, line 295: Will this also work with `.jpg` jpeg files?

imcgonigle (dismissed):
I've added a couple of questions but this looks good to me.

app/api/v0/consumers.rb
gileswells: Both `.jpg` and `.jpeg` have a content type of `image/jpeg`

app/api/v0/consumers.rb
```diff
@@ -285,6 +288,23 @@ def subscribed?(consumers)
                       kong_consumer: kong_consumer,
                       okta_consumers: okta_consumers
       end
+
```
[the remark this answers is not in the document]
gileswells: That would make sense.

config/environments/test.rb
```diff
@@ -95,4 +95,9 @@
   ENV['AUTHZ_SERVER_TEST'] = 't3st_4uthz'
   ENV['PROVIDER_ACG'] = 't3st_4uthz'
   ENV['PROVIDER_CCG'] = 't3st_4uthz'
+  ENV['S3_ACCESS_KEY_ID'] = 'AKIAQWERTYUIOPASAMPLE'
+  ENV['S3_LOGO_BUCKET_NAME'] = 'bucket-name'
+  ENV['S3_ROLE_ARN'] = 'arn:aws-us-gov:iam::000008675309:role/s3-upload-role'
+  ENV['S3_ROLE_SESSION_NAME'] = 'LPBLogoUpload'
+  ENV['S3_SECRET_ACCESS_KEY'] = 'TopSecret-qwertyuiopasdfghjklzxcvbnmqwer'
```
gileswells, line 102: If not abundantly clear, these are all dummy values, and additionally, real values were never committed to the repository.

jw81: Out of curiosity, if all these values were set to ridiculously obvious fake values, would it break anything?
Like, if they were changed to stuff like, `this_is_a_completely_bogus_value`, would everything still work?
Just thinking ahead for when/if a Fortify/WASA scan were to flag these, would it be immediately clear to everyone that these are obvious fake values.

app/views/production_mailer/support_production_access.html.erb
```diff
@@ -208,6 +208,16 @@
         <div><%= @request[:appDescription] %></div>
       </div>
     </li>
+    <% if @request[:logoLarge].present? %>
```
jw81, line 211: Do we need to check if both `@request[:logoLarge]` and `@request[:logoIcon]` are present?
Looks like we try to display `:logoIcon` and I'm just wondering if it's possible we could hit a scenario where `:logoLarge` was passed, but `:logoIcon` wasn't.

app/api/v0/consumers.rb
```diff
@@ -294,6 +297,30 @@ def subscribed?(consumers)
...
+                 regexp: { value: %r{^image/(jpeg|png)$},
+                           message: 'Files must be one of these types: image/jpeg, image/png' }
+      end
+      post 'logo-upload' do
+        header 'Access-Control-Allow-Origin', request.host_with_port
+        protect_from_forgery
+
+        aws = AwsSigv4Service.new
+        aws.set_key("#{SecureRandom.uuid}/#{params[:fileName]}")
+        aws.set_content_type(params[:fileType])
```
jw81, line 320: I'm surprised Rubocop doesn't bark about this. 🤔 
Usually, it'll tell you to favor `attr_writer` or `attr_accessor` declarations instead of building `set_blah()` methods. 🤷

app/api/v0/consumers.rb
```diff
@@ -294,6 +297,30 @@ def subscribed?(consumers)
...
+
+        aws.sign_request
```
jw81, line 322: It looks like most every endpoint that returns a response body, is first running the object through a Grape Entity facade.  i.e., `present my_cool_obj, with: V0::Entities::MyCoolObjEntity`.

Any value in maybe following that pattern here?

app/views/production_mailer/support_production_access.html.erb
gileswells: The only time that would happen is if there's some failure with the resizing that causes the large to succeed but the small one to fail. Chances of that seem insanely low. Both images are created through the same method and added to the post body at the same time.

app/api/v0/consumers.rb
gileswells: 🤷

app/api/v0/consumers.rb
gileswells: I don't see that being a bad addition. Can't hurt and likely is doing something under the hood more than what Rails does out of the box that's beneficial.

app/api/v0/entities/aws_sigv4_upload_entity.rb
```diff
@@ -0,0 +1,22 @@
...
+      expose :logoUrls, documentation: { is_array: true }
+      expose :policy, documentation: { type: String }
+      expose :resizeTriggerUrls, documentation: { is_array: true }
+      expose :s3_region_endpoint, documentation: { type: String }
+      expose :x_amz_algorithm, documentation: { type: String }
+      expose :x_amz_credential, documentation: { type: String }
+      expose :x_amz_date, documentation: { type: String }
+      expose :x_amz_expires, documentation: { type: String }
+      expose :x_amz_security_token, documentation: { type: String }
+      expose :x_amz_signature, documentation: { type: String }
```
jw81, line 19: Shoot I just thought of this.  Are all the attributes in the JSON response coming back in a `camelCase` format?
Looks like maybe some are in `camelCase` and some are in `snake_case`.  🤔

jw81 (approved):

PR #233, opened 2022-11-16
[API-21296] Update field descriptions to make it more clear

gileswells:
Despite my best efforts to make the Swagger placeholder text actually accurate, the end result was to give more emphasis on these fields needing to be comma separated. Not happy with this solution but based on research it seems this is not something that has been solved.

The root issue is that with using multiple lines, when you post sandbox and production the `api_environments_attributes[environments_attributes][name]` field it will show up twice without a trailing `[]` on it causing the backend to only receive the last value.

```
api_environments_attributes[environments_attributes][name]: sandbox
api_environments_attributes[environments_attributes][name]: production
```

### DIFF
--- a/app/api/v0/providers.rb
+++ b/app/api/v0/providers.rb
@@ -77,7 +77,7 @@ module V0
       desc 'Add an API to the developer portal' do
         detail <<-detailText
           # Environment and Scopes
-          These fields are CSV ignore what the placeholder text from the textarea says, it's wrong.
+          These fields are CSV. Ignore what the placeholder text from the textarea says. It's wrong.
 
           # Sample values (New Demo API)
           ## Basic details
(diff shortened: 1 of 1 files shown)


app/api/v0/providers.rb
```diff
@@ -74,51 +74,132 @@ def create_internal_consumer(user)
         present api, with: V0::Entities::ApiEntity
       end
 
+      desc 'Add an API to the developer portal' do
+        detail <<-detailText
+          # Environment and Scopes
+          These fields are CSV ignore what the placeholder text from the textarea says, it's wrong.
```
Cgfor3, line 80:
```suggestion
          These fields are CSV. Ignore what the placeholder text from the textarea says. It's wrong.
```

Cgfor3 (approved):
LGTM

PR #235, opened 2022-11-22
[API-14772] - Adds improved "auto deploy" GHA workflows for LPB

jw81:
Original Issue:: [API-14772](https://vajira.max.gov/browse/API-14772)

**Notes**
- When we are ready to officially turn on auto deploys, we'll need to go into the `lighthouse-devops-support` repo and create the following labels ::
  - `repo: lighthouse-platform-backend`
  - `MR in progress`

### DIFF
--- a/.github/workflows/auto-deploy-to-production.yml
+++ b/.github/workflows/auto-deploy-to-production.yml
@@ -6,18 +6,11 @@ on:
     - cron: '*/30 13,14,15,16,17,18,19,20,21,22,23 * * 1,2,3,4,5'
 
 env:
-  environment: 'QA'
+  environment: 'Production'
 
 jobs:
   run_checks_and_deploy:
     runs-on: ubuntu-latest
-    outputs:
-      time_string: ${{ steps.get_time_string.outputs.time_string }}
-      mr_number: ${{ steps.get_mr_number.outputs.mr_number }}
-      build_number: ${{ steps.deploy_to_production.outputs.build_number }}
-      start_time: ${{ steps.check_build_status.outputs.start_time }}
-      end_time: ${{ steps.check_build_status.outputs.end_time }}
-      execution: ${{ steps.check_build_status.outputs.execution }}
     steps:
       - name: Set timezone to Eastern
         uses: szenius/set-timezone@v1.0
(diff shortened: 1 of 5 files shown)


.github/workflows/auto-deploy-to-production.yml
```diff
@@ -27,57 +20,55 @@ jobs:
...
         run: |
           TIME_STRING=`echo ${{steps.get_time_string.outputs.time_string}}`
-          echo $TIME_STRING
-          echo ${{ secrets.GITHUB_TOKEN }} | gh auth login --with-token
+          echo "The time is: $TIME_STRING"
+          echo ${{ secrets.GIT_AUTO_DEPLOY_TOKEN }} | gh auth login --with-token
           gh issue list \
-            -R 'department-of-veterans-affairs/lighthouse-platform-backend' \
+            -R 'department-of-veterans-affairs/lighthouse-devops-support' \
             -l 'repo: lighthouse-platform-backend' \
```
gileswells, line 33: I created the label over on the repo just so we don't miss it and bork things up

.github/workflows/auto-deploy-to-production.yml
```diff
@@ -27,57 +20,55 @@ jobs:
...
             --limit 1 \
             --json id,number,state,labels,title,body \
             --jq "[.[]|select(.title==\"Deploy lighthouse-platform-backend to production ($TIME_STRING)\")][0]" \
             > issue.json
           cat issue.json
-          MR_NUMBER=`jq -e ".number" issue.json`
-          echo $MR_NUMBER
+          MR_NUMBER=`jq ".number" issue.json`
```
gileswells, line 39: I have this same line including the `-e` on the developer-portal auto deploy script. Is this not needed and simply an extra flag?

.github/workflows/auto-deploy-to-production.yml
```diff
@@ -27,57 +20,55 @@ jobs:
...
-          echo "::set-output name=mr_number::$MR_NUMBER"
+          echo "mr_number=$MR_NUMBER" >> $GITHUB_OUTPUT
       - id: deploy_to_production
         name: Deploy to production
         if: steps.get_mr_number.outcome == 'success'
         continue-on-error: true
         run: |
           echo "pre-fetching next build number..."
-          NEXT_BUILD=`curl -u ${{ secrets.user_jenkins }}:${{ secrets.api_token_jenkins }} https://tools.health.dev-developer.va.gov/jenkins/job/department-of-veterans-affairs/job/health-apis-deployer/job/d2/api/json | jq '.nextBuildNumber'
+          NEXT_BUILD=`curl -u ${{ secrets.user_jenkins }}:${{ secrets.api_token_jenkins }} https://tools.health.dev-developer.va.gov/jenkins/job/department-of-veterans-affairs/job/health-apis-deployer/job/d2/api/json | jq '.nextBuildNumber'`
```
gileswells, line 58: This causes a potential race condition. When we queue a build in the next curl command we get a response that includes the queue item. Could we poll that item until it returns the build number instead? I don't think this would be a big chance of failure but I think it might be best to ensure we're getting the right build job to determine our success / failure and duration information from.

.github/workflows/auto-deploy-to-production.yml
```diff
@@ -87,15 +78,17 @@ jobs:
...
+            echo "The status is ... $STATUS"
             if [ $STATUS -ne 200 ]; then
               echo 'Failed to locate build...'
               echo 'exiting...'
               exit 1;
             fi
-            USER_ID=`jq '.actions[] | select(._class == "hudson.model.CauseAction") | .causes[0].userId' build.json`
-            if [ "$USER_ID" == "okapibot" ]; then
+            USER_ID=`jq '.actions[] | select(._class == "hudson.model.CauseAction") | .causes[0].userId' build.json | xargs echo -n`
+            if [ "$USER_ID" = "okapibot" ]; then
```
gileswells, line 90: 🤔 Looks like you've got a catch for after the race condition to cure itself if needed. This appears to be of sound logic. Maybe add an echo showing we hit the race condition before doing the `BUILD_NUMBER++` just for posterity?

.github/workflows/auto-deploy-to-production.yml
```diff
@@ -104,30 +97,28 @@ jobs:
...
               echo 'Oh noes, something borked...'
-              echo "https://tools.health.dev-developer.va.gov/jenkins/job/department-of-veterans-affairs/job/health-apis-deployer/job/d2/$build_number"
+              echo "https://tools.health.dev-developer.va.gov/jenkins/job/department-of-veterans-affairs/job/health-apis-deployer/job/d2/$BUILD_NUMBER"
               exit 1;
             fi
-            if [ "$STATUS" == "SUCCESS" ]
+            if [ "$STATUS" = "SUCCESS" ]; then
               RUNNING="false"
             else
+              echo "Job status is :: $STATUS.  Sleeping 60 seconds..."
```
gileswells, line 109: This could potentially provide an infinite loop. Without a condition that catches a status of FAILURE or any of the other final and failing build statuses this could go on forever. (this is also something that we should update on the dev portal auto-deploy)

Technically not broken so not a blocker but something to consider perhaps.

gileswells (approved):
I do have some questions on some things and a potential improvement for the future (re: infinite loop) but overall I think this is mergable as is.

I'm not the largest fan of each merge to master triggers a new MR. I feel like we may have instances where we need to merge multiple PRs into a single release and this flow, as it stands right now, would require that we merge those to a combining branch before we could merge to master.

.github/workflows/auto-deploy-to-production.yml
gileswells: _see next comment_

.github/workflows/auto-deploy-to-production.yml
jw81: TY! 🥳

jw81:
> I do have some questions on some things and a potential improvement for the future (re: infinite loop) but overall I think this is mergable as is.
> 
> I'm not the largest fan of each merge to master triggers a new MR. I feel like we may have instances where we need to merge multiple PRs into a single release and this flow, as it stands right now, would require that we merge those to a combining branch before we could merge to master.

Alex had mentioned to me (and I agree) that his original thinking was that this is really a first iteration.  Something that will likely need to be finessed as we go along.  So I'm totally open to tweaking/enhancing it as we go.

PR #241, opened 2022-12-22
Migrate ACG/CCG audience values to LPB

andresriveratoro:
https://vajira.max.gov/browse/API-21234


spec/factories/api_metadata.rb
```diff
@@ -13,6 +13,8 @@
           baseAuthPath: "/oauth2/#{Faker::Lorem.word}/v1",
           productionWellKnownConfig: Faker::Internet.url,
           sandboxWellKnownConfig: Faker::Internet.url,
+          productionAud: Faker::Internet.slug,
+          sandboxAud: Faker::Internet.slug,
```
jw81, line 17: Any value in making both of these factories use the same kind of fake data?
This factory is setting the audience values via `Faker::Internet.slug`, but the provider factory is setting the audience values via `Faker::Lorem.word`.

lib/tasks/aud_values.rake
```diff
@@ -0,0 +1,99 @@
...
+            'sandboxAud' => 'ausftw7zk6eHr7gMN2p7',
+            'productionAud' => 'ausftw7zk6eHr7gMN2p7'
+          }
+        },
+        'veteran_verification' => {
+          'acgInfo' => {
+            'sandboxAud' => 'aus7y0sefudDrg2HI2p7',
+            'productionAud' => 'aus7y0sefudDrg2HI2p7'
+          }
+        }
```
jw81, line 88: @gileswells @andresriveratoro 
I wanna make sure I understand what's happening here.  So please correct me if needed.

So today, when a new Lighthouse API is being onboarded, we have to add that new API to LPB.
One of the attributes required for creating an API are these audience values.
Those audience values are given to us by someone (maybe Pivot?).
We store those audience values in the AWS Params Store under a name like, `/dvp/{environment}/platform-backend/authz_server_claims` and `/dvp/{environment}/platform-backend/authz_server_claims_ccg`.
When we create the API, the value we're providing for the `auth_server_access_key` attribute on an API model is the name of that params store key.  For example, `authz_server_claims`.
So when we need to answer the question of, "what auth types does this API allow?", we just look for the presence of a value in the params store.

With this change, we're now just going to take that audience value we received (from Pivot?) and store it in the `oauth_info` attribute of the API's related metadata record.  Then we can check the metadata record to know what auth types the API provides.

Is my understanding correct, or am I off?

spec/factories/api_metadata.rb
andresriveratoro: I went with these values based on the [existing code](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/master/spec/factories/api_metadata.rb#L20), but I will happily make the change for consistency. They both return strings.

https://www.rubydoc.info/gems/faker/Faker%2FLorem.word
https://www.rubydoc.info/gems/faker/Faker%2FInternet.slug

andresriveratoro: Let me know if you would like me to update them or keep as is.

lib/tasks/aud_values.rake
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+namespace :aud_values do
+    desc 'inject OAuth audience values'
```
gileswells, line 4: Indent here and all the way down from here appears wrong. Shouldn't this be 2 spaces?

lib/tasks/aud_values.rake
gileswells: Some of the technical steps are missed in your larger explanation of how it works now but essentially you're correct. 

On build, Jenkins grabs the paramstore values and sets them to environment values in the server from [this file](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend-deployment/blob/master/ecs/ecs-params.yml). Because it's not a live fetch from Paramstore any new API being added needs a fresh deployment to all 3 environments.

Since these values are public facing and have zero security risk there doesn't appear to be any reason to store these in Paramstore at all.

Also this is a one time job. This file will not need to be updated over time with new audience values. Those will come as we create new ones through the Swagger UI posting that we use to create new APIs.

db/migrate/20230125163247_add_audience_values.rb
```diff
@@ -0,0 +1,5 @@
+class AddAudienceValues < ActiveRecord::Migration[7.0]
+  def up
+    Rake::Task['aud_values:seedAudienceValues'].invoke
```
gileswells, line 3: I think I'd prefer to have this included within the `lpb` namespace. Easier to bundle all of our custom stuff into one place so we always know where it'd be found under the list of `rake` tasks.

lib/tasks/aud_values.rake
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+namespace :aud_values do
```
gileswells, line 3: Unsure if you can simply change this to `:lpb` for the namespace to bring it together with the well known configs task or if you'd need to move this code to the other `lib/tasks/lpb.rake` file. If changing just the namespace works then I'm totally fine with having two different files that have tasks which add to the `lpb` namespace.

rakelib/ci.rake
```diff
@@ -16,8 +16,8 @@ namespace :spec do
       Rake::Task['spec'].invoke
     rescue SystemExit => e
       e.status
-                  else
-                    0
+    else
+      0
```
gileswells, line 20: I didn't dig into it further but I ran into `rubocop` not liking fixing this indentation. Not sure why or how the test is failing but you might have issues with this change and the Jenkins tests. If this works just fine then feel free to ignore this comment entirely.

lib/tasks/aud_values.rake
```diff
@@ -0,0 +1,99 @@
...
+          }
+        }
+      }.freeze
+      audience_values[url_fragment][type][audience]
+    end
+  
+    def update_oauth_aud_values(api, oauth_type)
+      url_fragment = api[:url_fragment]
+      api.oauth_info[oauth_type]['productionAud'] = get_aud_values(url_fragment, oauth_type, 'productionAud')
+      api.oauth_info[oauth_type]['sandboxAud'] = get_aud_values(url_fragment, oauth_type, 'sandboxAud')
```
gileswells, line 96: These values should both equal the `sandboxAud` value in the above config object unless the environment is production where in that case we should have `sandboxAud = sandboxAud` and `productionAud = productionAud`.

Also with this change we do need to have the proper `productionAud` values added to the config above.

gileswells (changes requested):
Some updates needed to make sure that the seed into production works as well as some other cleanup items.

spec/factories/api_metadata.rb
jw81: Ah, I see what you're saying now.  Each factory was already using a faker method and you just kept it consistent with what was already there.  Works for me.  👍

rakelib/ci.rake
andresriveratoro: I've reverted the change. I'm not sure how that got there.

db/migrate/20230125163247_add_audience_values.rb
andresriveratoro: I've added it to the `lpb` namespace 👍

PR #243, opened 2023-01-03
[API-14772] - add logic to re-login with the devops creds

jw81:
[Original Issue](https://vajira.max.gov/browse/API-14772)

When we first tried to run the auto-deploy workflows against the DevOps repo, we encountered the following error in the `create_mr` step of the `Create Auto Deploy Maintenance Request` workflow ::
```
GraphQL: Could not resolve to a Repository with the name 'department-of-veterans-affairs/lighthouse-devops-support'. (repository)
```

Looks like we might need to re-login using the DevOps github creds  in that step.  This PR adds that re-login functionality.

### DIFF
--- a/.github/workflows/create-auto-deploy-maintenance-request.yml
+++ b/.github/workflows/create-auto-deploy-maintenance-request.yml
@@ -35,8 +35,6 @@ jobs:
       - id: create_mr
         name: Generate the MR body text and create MR
         if: steps.check_for_existing_mr.outcome == 'success'
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/create-auto-deploy-maintenance-request.yml
```diff
@@ -35,8 +35,6 @@ jobs:
       - id: create_mr
         name: Generate the MR body text and create MR
         if: steps.check_for_existing_mr.outcome == 'success'
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
gileswells, line 39: This token is generated per workflow run and only has repository access so this change makes sense.

gileswells (approved):

PR #245, opened 2023-01-09
[API-22785] - Add well known configs

gileswells:
https://vajira.max.gov/browse/API-22785

This adds support for the .well-known config urls to be added to the oauth_info field. This is in support of automation around API scopes.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,7 +5,7 @@ values =  HTTParty.get('https://urlhaus.abuse.ch/downloads/text/').body
                                                                   .split("\r\n")
                                                                   .delete_if { |line| line.starts_with?('#') }
                                                                   .map { |line| [line] }
-MaliciousUrl.import columns, values, validate: false, on_duplicate_key_ignore: true
+# MaliciousUrl.import columns, values, validate: false, on_duplicate_key_ignore: true
 
 
 appeals_category = ApiCategory.create(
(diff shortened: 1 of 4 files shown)


db/seeds.rb
```diff
@@ -5,7 +5,7 @@
                                                                   .split("\r\n")
                                                                   .delete_if { |line| line.starts_with?('#') }
                                                                   .map { |line| [line] }
-MaliciousUrl.import columns, values, validate: false, on_duplicate_key_ignore: true
+# MaliciousUrl.import columns, values, validate: false, on_duplicate_key_ignore: true
```
jw81, line 8: Do we not want this anymore?

gileswells: Just noticed that locally. Was an unintentional commit.
```suggestion
MaliciousUrl.import columns, values, validate: false, on_duplicate_key_ignore: true
```

gileswells: Commenting this line takes the `rake db:seed` step of rebuilding the app from more than 10-15 minutes to just about 30 seconds.

jw81:
Any value in maybe adding a simple test to make sure that a created API has this new attribute structure as you'd expect?  Maybe somewhere in [this area](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/master/spec/api/v0/providers_spec.rb#L103)?  (totally cool btw to not agree.  just wanted to see if you think there's any value there)

jw81 (approved):

PR #246, opened 2023-01-18
[API-22785] - Seed well known configs

gileswells:
https://vajira.max.gov/browse/API-22785

Seed the well-known config urls to the database using a migration and rake task. This prevents having to add these potentially breaking data changes manually in production.

### DIFF
--- a/lib/tasks/lpb.rake
+++ b/lib/tasks/lpb.rake
@@ -1,0 +1,82 @@
+namespace :lpb do
+  desc "Inject sandbox and production .well-known config urls"
+  task seedWellKnownConfigValues: :environment do
+    apis = ApiMetadatum.where.not(oauth_info: nil)
+    # apis = [ApiMetadatum.find(3)]
+    p apis.length
+    apis.each do |api|
+      p api[:display_name]
+      if api.oauth_info['acgInfo'].present?
+        process_auth_type(api, 'acgInfo')
+      end
+      if api.oauth_info['ccgInfo'].present?
+        process_auth_type(api, 'ccgInfo')
+      end
+    end
+  end
+
+  def process_auth_type(api, type)
+    configBaseUrl = {
+      'development' => 'https://dev-api.va.gov',
+      'production' => 'https://api.va.gov',
+      'sandbox' => 'https://sandbox-api.va.gov',
+      'staging' => 'https://staging-api.va.gov',
+      'test' => 'https://dev-api.va.gov'
+    }.freeze
+    config = { 
+      'claims' => {
+        'acgInfo' => '/oauth2/claims/v1/.well-known/openid-configuration', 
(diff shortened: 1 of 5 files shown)


andresriveratoro (approved):
LGTM

lib/tasks/lpb.rake
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+
+namespace :lpb do
+  desc 'Inject sandbox and production .well-known config urls'
+  task seedWellKnownConfigValues: :environment do
+    apis = ApiMetadatum.where.not(oauth_info: nil)
+    # apis = [ApiMetadatum.find(3)]
```
jw81, line 7: Still need this?

lib/tasks/lpb.rake
```diff
@@ -0,0 +1,84 @@
...
+      'pgd' => {
+        'ccgInfo' => '/oauth2/pgd/system/v1/.well-known/openid-configuration'
+      }
+    }.freeze
+    paths[url_fragment][type]
+  end
+
+  def process_auth_type(api, type)
+    url_fragment = api[:url_fragment]
+    environment = ENV.fetch('RAILS_ENV')
```
jw81, line 76: Are we certain that the `RAILS_ENV` variable evaluates to what we expect in all the different environments?
I know on past projects, that environment variable will actually evaluate to "production" even when it's really running in a Sandbox environment.  Pretty sure that's what it does in `vets-api`.

lib/tasks/lpb.rake
```diff
@@ -0,0 +1,84 @@
...
+      process_auth_type(api, 'ccgInfo') if api.oauth_info['ccgInfo'].present?
+    end
+  end
+
+  def get_base_url(environment)
+    base_urls = {
+      'development' => 'https://dev-api.va.gov',
+      'production' => 'https://api.va.gov',
+      'sandbox' => 'https://sandbox-api.va.gov',
+      'staging' => 'https://staging-api.va.gov',
```
jw81, line 21: Related to [this comment](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/246/files#r1084192200), I could always be wrong, but I don't think we're ever actually going to see `RAILS_ENV` evaluate to sandbox or staging.  I think it can only evaluate to production, development, or test.

gileswells: The reason for Sandbox is around needing two different urls for the production LPB instance. I do agree that we'll never see a true `sandbox` environment but just needed to fudge it here for the different hostnames for sandbox and production configs.

lib/tasks/lpb.rake
gileswells: no

lib/tasks/lpb.rake
gileswells: fixed

jw81 (approved):

PR #255, opened 2023-02-13
Add news to LPB

andresriveratoro:
https://vajira.max.gov/browse/API-22680


db/migrate/20230213145053_create_news.rb
```diff
@@ -0,0 +1,11 @@
+class CreateNews < ActiveRecord::Migration[7.0]
+  def change
+    create_table :news do |t|
+      t.string :title
+      t.string :url
+      t.string :category
+
```
gileswells, line 7: Don't you need a `source` column as well?

app/models/news.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class News < ApplicationRecord
+  include Discard::Model
+
+  CATEGORIES = {
+    articles: 'articles',
+    digital_media: 'digital_media',
+    news_releases: 'news_releases'
+  }.freeze
```
gileswells, line 10: The ticket calls out adding news and categories to the database. To me, I would much prefer to see the categories added as a related table with fields that cover the config and on page text shows in the developer portal that is currently coming from [this file](https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/content/news.yml).

gileswells (changes requested):
Good first pass but I think we can get in some additional improvements while we're kinda holding for webpack / nodejs updates prior to IA work.

andresriveratoro:
Closing in favor of https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/259

PR #256, opened 2023-02-13
[API-24249] - Add a rescue for fields with no parsable json

gileswells:
https://vajira.max.gov/browse/API-24249

Provide a fallback for fields that don't already have a black JSON compliant value of `{}`. This could be accomplished by setting this value with a migration / task but wanted to do it at runtime to allow for newly created rows to still work if they somehow bypassed the default value setup of `{}`.


Cgfor3 (approved):
Looks good :shipit:

PR #258, opened 2023-02-17
[API-24383] - Update how credentials payload key is constructed

gileswells:
https://vajira.max.gov/browse/API-24383

Oauth ACG signups aren't properly setting the credentials property in the Okta payload. This is a legacy bug from the migration form `developer-portal-backend` to `lighthouse-platform-backend` which was recently uncovered.

### DIFF
--- a/app/services/okta/base_service.rb
+++ b/app/services/okta/base_service.rb
@@ -138,6 +138,17 @@ module Okta
       "#{consumer_name(user)}-#{Time.now.to_i}#{lower_env? ? '-dev' : ''}"
     end
 
+    def construct_credentials(application_type)
+      if %w[web service].include? application_type
+        { oauthClient: {
+          autoKeyRotation: true,
+          token_endpoint_auth_method: application_type == 'web' ? 'client_secret_basic' : 'private_key_jwt'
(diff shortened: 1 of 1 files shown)


jw81 (approved):

app/services/okta/base_service.rb
```diff
@@ -138,6 +138,17 @@ def construct_label(user)
       "#{consumer_name(user)}-#{Time.now.to_i}#{lower_env? ? '-dev' : ''}"
     end
 
+    def construct_credentials(application_type)
```
jw81, line 141: Just double-checking, but we're guaranteed that `application_type` will always be present and will always be a value of `web`, `service`, or `native`?

gileswells: It is an optional field for the actual post request from dev portal an ACG signup `key: oAuthApplicationType, values: ['web', 'native']`. That being said, if it gets to this point in the flow then it has already been determined that it is at least an oauth signup request and if it's a CCG signup request it will hard set it as `service` when calling the `construct_credentials` method.

andresriveratoro (approved):

PR #259, opened 2023-02-20
Add news to LPB

andresriveratoro:
https://vajira.max.gov/browse/API-22680


app/api/v0/news.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
```
jw81, line 1: The structures of these URLs are interesting.
It looks like we'd be adding these new URLs ::
```
GET /v0/news (returns all NewsCategory records)
POST /v0/news/category (creates a new NewsCategory record)
POST /v0/news/item (creates a new NewsItem record for a related NewsCategory)
```

I guess I would maybe expect the URLs to look a bit more "RESTful".  Something more like this ::
```
GET /v0/news
POST /v0/news/categories (pluralized form of the resource)
POST /v0/news/categories/{category_identifier}/items (pluralized form of the resource and it's explicitly tied to a specific NewsCategory instance)
```

Definitely curious to see what you and @gileswells (and anyone else) think.

spec/api/v0/news_spec.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
```
jw81, line 1: Looks like we're also adding 2 new POST methods.  We'd def want to add some specs for those as well.

jw81 (changes requested):
I think there's probably some specs we still need to add for the new POST urls and there's prob some discussion to have around the structuring of the new URLs.  I left some comments in the PR around both topics.

jw81 (approved):

PR #264, opened 2023-03-27
API-24812 add okta app profile label

johnmweisz:
will add profile.label attribute to new Okta apps
https://vajira.max.gov/browse/API-24812


dbrown-va (approved):
This adds functionality to support a new access token claim being pursued for a user friendly App Name rather than the Okta client id.

The claim minting will be enabled separately under a future MR.

Pivot! will perform the existing client backfilling.

jw81 (approved):

PR #268, opened 2023-03-30
[API-23993] - Schema updates to support DevPortal IA changes

jw81:
Original Issue :: [API-23993](https://vajira.max.gov/browse/API-23993)

Adds the following columns to the `api_metadata` table
- `overview_page_content`
- `restricted_access_details`
- `restricted_access_toggle`
- `url_slug`

Adds the following column to the `api_categories` table
- `url_slug`


db/migrate/20230330194428_add_overview_page_content_to_api_metadata.rb
```diff
@@ -0,0 +1,5 @@
+class AddOverviewPageContentToApiMetadata < ActiveRecord::Migration[7.0]
+  def change
+    add_column :api_metadata, :overview_page_content, :text
```
gileswells, line 3: Any insights to be gained around `rails_admin` support for fields containing markdown?

gileswells (approved):
Why on earth did you do this as 5 migrations as opposed to 1? I mean it works so ✅ but seems odd. Had one question about handling markdown within `rails_admin`.

jw81:
> Why on earth did you do this as 5 migrations as opposed to 1? I mean it works so ✅ but seems odd. Had one question about handling markdown within `rails_admin`.

Haha, I like using the rails generators and AFAIK, the generators won't let you do things like add columns to multiple tables in a single shot.

PR #281, opened 2023-05-25
API-23996 Update Developer Portal to use url_slug

imcgonigle:
[Jira Ticket](https://vajira.max.gov/browse/API-23996?filter=-1)

This PR updates the seed data so that url-slugs are unique instead of  "default-slug". This is necessary for the Developer Portal function as expected.


Cgfor3 (approved):
:shipit:

gileswells (approved):

PR #287, opened 2023-06-30
Api 23999/api overview page

imcgonigle:
Tis PR finishes up API 23999 with the following changes.
- Adds missing "Getting access" content
- Fixes the API titles being used
- Adds the following new APIs
> - Notice of Disagreements
> - Supplemental Claims
> - Legacy Appeals API
> - Higher Level Reviews API
> - Appealable Issues API


jw81 (approved):
LGTM 👍

PR #288, opened 2023-07-01
Fix category url_slug fields

gileswells:
The logic in the `lpb:seedIAFieldValues` rake task for creating category url_slug values is wrong. This resolves that.


jw81 (approved):
LGTM 👍

PR #292, opened 2023-07-03
Update Higher Level Reviews in overview content rake task

imcgonigle:
This PR fixes a mistake I made in a previous PR.

In a earlier PR I was trying to change 'Higher Level Reviews API' to 'Higher-Level Reviews API' but I mistakenly changed the switch case to check for the updated name ("Higher-Level Reviews API"), which wasn't in the DB yet, instead of what was in the DB ("Higher Level Reviews API")


Cgfor3 (approved):
:shipit:

PR #301, opened 2023-07-19
[API-27989] - Enhance Security Alert GHA Workflow

jw81:
Original Issue :: [API-27989](https://jira.devops.va.gov/browse/API-27989)

- Adds logic to check for open Dependabot alerts
- Adds logic to check for open Secret Scanning alerts
- Adds logic to send a message to the ~`#team-okapi`~ `#team-okapi-alerts` Lighthouse Slack channel if an open alert is found


.github/workflows/notify-security-alerts.yml
```diff
@@ -20,8 +20,59 @@ jobs:
...
+        uses: slackapi/slack-github-action@v1.23.0
+        if: |
+          steps.check_for_code_scanning_alerts.outputs.security_alert_exists == 'true' ||
+          steps.check_for_dependabot_alerts.outputs.security_alert_exists == 'true' ||
+          steps.check_for_secret_scanning_alerts.outputs.security_alert_exists == 'true'
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_API_TOKEN }}
+        with:
+          # This ID is for #team-okapi in Lighthouse Slack
+          channel-id: C01931CFMTQ
```
gileswells, line 58: I feel like we should create a #team-okapi-alerts channel rather than shoot this into our default channel. If there is something that we're seeing come up and it takes us a few days to actually be able to get to it we're going to have this post multiple times. Might make it a bit cluttered in a channel for collaboration.

gileswells (changes requested):
Consider changing the channel target.

.github/workflows/notify-security-alerts.yml
jw81: Ah, I did that because it was requested that all teams set this up to report to their respective `#team-X` channels in Lighthouse Slack.  

> _In further support of this ask, please set up a Slack alert which reports these to your #team- channel for all of your GitHub repositories._

[Reference](https://lighthouseva.slack.com/archives/CAZF880HW/p1689004091121079)

gileswells: I get that but I still think that `#team-okapi-alerts` is a fair way to do it especially if we make note in our main channel the presence of the second channel and cross reference an alert if it's something that's going to require a larger effort.

Like, if we manage it well, it should be fine. I'd prefer to go with an alerts channel but if they give push back then we can move it. More a preference but I know that this is going to be to satisfy an external requirement so it's more so up to them.

jw81: Fair enough.  We can try spinning up an alerts channel and sending the alerts there.  We can always point it back to our main channel if they really insist upon that.

jw81: Created the new `#team-okapi-alerts` Slack channel and changed the workflow to send to that new channel.

gileswells (approved):

PR #304, opened 2023-07-25
[API-28380] - remove old codeql configuration

jw81:
Original Issue :: [API-28380](https://jira.devops.va.gov/browse/API-28380)

- removes old Codeql configuration so it can be replaced by the new configuration
  - A [previous PR](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/282) attempted to do this in one shot, but that build won't go green because it references back to the old CodeQl config that lives in Master.  So let's just remove this old config and immediately replace it with the new one in a followup PR.


Cgfor3 (approved):
👍

PR #307, opened 2023-07-27
[API-28463] - Update MR template

gileswells:
https://jira.devops.va.gov/browse/API-28463


gileswells:
Closing for the other one to be used.

PR #318, opened 2023-10-05
[API-23292] - Add deactivation_info conditional and test

gileswells:
https://jira.devops.va.gov/browse/API-23292

Update sitemap to block APIs that are deactivated and add swagger docs url to all APIs in the sitemap.


jw81 (approved):
LGTM 👍

PR #319, opened 2023-10-17
[API-30848] - Add is_stealth_launched boolean field

gileswells:
https://jira.devops.va.gov/browse/API-30848

Adds `is_stealth_launched` field to the `api_metadata` database table and exposes it through `isStealthLaunched` on the `legacy.json` API endpoint. The reason for the two migrations was to first default the field to false (which all current APIs should have) and then set it to true so that all future APIs will start as being stealth launched so Okapi can audit newly added APIs before making them visible to the public.

I also removed the rake tasks that we were using in migrations as they were causing issues when spinning up a new instance of the app. In the future we shouldn't use this pattern of calling a rake task within a migration. (That was my bad idea and this is me realizing and reversing it)

### DIFF
--- a/app/api/v0/entities/legacy/api_provider_entity.rb
+++ b/app/api/v0/entities/legacy/api_provider_entity.rb
@@ -79,6 +79,7 @@ module V0
         expose :restricted_access_details, as: :restrictedAccessDetails
         expose :restricted_access_toggle, as: :restrictedAccessToggle
         expose :url_slug, as: :urlSlug
+        expose :is_stealth_launched, as: :isStealthLaunched
       end
     end
   end
--- a/db/migrate/20230111172809_add_well_known_configs.rb
+++ b/db/migrate/20230111172809_add_well_known_configs.rb
@@ -1,5 +1,5 @@
 class AddWellKnownConfigs < ActiveRecord::Migration[7.0]
   def up
-    Rake::Task['lpb:seedWellKnownConfigValues'].invoke
+    # Rake::Task['lpb:seedWellKnownConfigValues'].invoke
   end
 end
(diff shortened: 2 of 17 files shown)


jw81:
> The reason for the two migrations was to first default the field to false 

Are we 100% sure this is how ActiveRecord behaves?
I wanna say that I think this won't go back and set the new column value for existing records.  I'm pretty sure it only affects records that are created after the migration is ran.  (I could absolutely be wrong though).

gileswells:
> I wanna say that I think this won't go back and set the new column value for existing records.

That's correct and what we want to happen. The first migration to add the column, when you specify a default, will set that field to the default value (`false`) which is what we want all existing records to have. The second migration will set the default value for all _NEW_ records to `true` which allows us to go and check the APIs before we manually change it to `false` and allow it to show up on the Explore page on the site.

I just re-confirmed this on my local by adding a new API and it was added to the database with `is_shadow_launched` as `true` like is expected.

jw81:
> > I wanna say that I think this won't go back and set the new column value for existing records.
> 
> That's correct and what we want to happen. The first migration to add the column, when you specify a default, will set that field to the default value (`false`) which is what we want all existing records to have. The second migration will set the default value for all _NEW_ records to `true` which allows us to go and check the APIs before we manually change it to `false` and allow it to show up on the Explore page on the site.
> 
> I just re-confirmed this on my local by adding a new API and it was added to the database with `is_shadow_launched` as `true` like is expected.

I stand corrected!  I went and tried it out locally with some pre-existing records in the table and sure enough, that first migration DOES go back and provide the `false` value for the pre-existing records.  I thought for sure that it wouldn't, but I was wrong.  🤷

PR #323, opened 2023-12-12
[API-31636] - Support test user data deeplink

gileswells:
https://jira.devops.va.gov/browse/API-31636

This endpoint validates the request for test user data and then fetches it from s3 and provides it to the user.


app/api/v0/consumers.rb
```diff
@@ -147,17 +151,53 @@ def subscribed?(consumers)
         protect_from_forgery
 
         user = user_from_signup_params
+        deeplink_hash = ''
+        deeplink_url = ''
+        if 'oauth/acg'.in? params[:apis].first.locate_auth_types
+          url_slug = params[:apis].first.api_metadatum.url_slug
+          deeplink_hash = generate_deeplink_hash(user)
```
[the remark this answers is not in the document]
gileswells: I have this here for debugging purposes. Based on a flipper setting we can turn it on in dev where the sandbox form will return the hash so we can inspect the network traffic and validate the deeplink is working without turning on emails sending out from signups to dev.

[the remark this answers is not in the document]
jw81: Ahhhhhh, nvm.
I totally missed line#172 where you're referencing the `deeplink_hash` var.  
Ignore me.

app/api/v0/consumers.rb
```diff
@@ -147,17 +150,55 @@ def subscribed?(consumers)
...
+          deeplink_url = generate_deeplink(url_slug, user)
```
jw81, line 158: I think there might be a hiccup here.

I was running this locally and it looks like the `user` isn't yet saved at this point.  Meaning that when it gets into `generate_deeplink()`, the `user.id` is going to be `nil`.

I'm not sure exactly "why" it was setup this way, but it looks like the `User` is not saved until OKTA has had a chance to create the application. [REF](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/blob/master/app/services/okta/base_service.rb#L81)

Maybe the fix is just simply calling `user.save!` before hitting your new logic?  (trying to think through if there's any drawbacks/concerns of doing that 🤔 )

jw81 (changes requested):
Pointed out a slight hiccup that I think we'll need to account for.  Hopefully should be an easy fix.

andresriveratoro (approved):
LGTM

app/api/v0/consumers.rb
gileswells: I was resistant to this change but I realized that the user, on their first ever signup, would be nil at this point because they don't already exist like my test user does in my local. 🤦‍♂️ I've added the user.save! above inside this conditional so it'll only fire if we need the user.id at this point.

andresriveratoro (approved):

spec/spec_helper.rb
```diff
@@ -19,7 +19,7 @@
   add_filter '/lib/tasks/lpb.rake'
   add_filter '/lib/tasks/aud_values.rake'
 
-  SimpleCov.minimum_coverage_by_file 90
+  SimpleCov.minimum_coverage_by_file 85
```
gileswells, line 22: Ticket to re-raise the minimum coverage level
https://jira.devops.va.gov/browse/API-32847

app/api/v0/consumers.rb
jw81: I think it's prob fine like this, but I don't think the "user.save!" is necessary here because the 2 calls to ".consumer_signup()" up above would have already called "user.save!" before this.

PR #325, opened 2024-01-10
[API-32038] Add address validation endpoint

andresriveratoro:


app/api/utilities.rb
```diff
@@ -130,5 +132,50 @@ def okta_signup_options
...
+          end
+        end
+      end
+      post '/candidate' do
+        header 'Access-Control-Allow-Origin', request.host_with_port
+        protect_from_forgery
+
+        if Flipper.enabled? :address_validation
+          address_data = JSON.parse(request.body.read)
+          uri = URI.parse(ENV.fetch('ADDRESS_VALIDATION_API_V2_CANDIDATE_ENDPOINT'))
```
jw81, line 164: I wonder if we should go ahead and add this env var to the [ECS params](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend-deployment/blob/master/ecs/ecs-params.yml) now, even though we're not planning on using this endpoint any time soon.  🤔 

Thoughts?

andresriveratoro:
The need for address validation is no longer required.

PR #326, opened 2024-01-17
[API-31636] - Use AWS gem in ECS

gileswells:
https://jira.devops.va.gov/browse/API-31636

Updates how we fetch the test users file when we're hosted in the ECS environment where credentials to allow retrieval of the S3 hosted file are available through the ECS Credentials method of the AWS gem.

I mostly followed [this guide](https://aws.amazon.com/blogs/compute/a-guide-to-locally-testing-containers-with-amazon-ecs-local-endpoints-and-docker-compose/) and made tweaks to fit our needs.


app/services/aws_s3_service.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'securerandom'
```
jw81, line 3: Is this actually being used in this service class?

gileswells:
Looks like falling back to a simple GET request, while functional, is going to cause the file coverage to drop because the ECS authenticated flow in production env builds (for dev, staging, and prod environments) will never be run and drops the percentage very low. 

Working to implement this [ECS authentication mock server](https://aws.amazon.com/blogs/compute/a-guide-to-locally-testing-containers-with-amazon-ecs-local-endpoints-and-docker-compose/) method.

jw81 (approved):
👍

andresriveratoro (approved):

PR #327, opened 2024-01-19
[API-27141] - Add company address to production requests

andresriveratoro:
Part of https://jira.devops.va.gov/browse/API-27141


jw81:
Has the frontend already been updated/deployed to begin collecting these new `Address` fields and sending that data to LPB?

I only ask because if the frontend isn't sending this data yet, then when we deploy this code, we'll start seeing failures on the Prod Access form because there will suddenly be `required` data attributes in LPB that aren't present.

jw81 (approved):
I ran this locally against the Master branch of the DevPortal and everything seems to be functioning as expected.  👍

PR #328, opened 2024-01-20
Use single source of truth for Ruby

andresriveratoro:
This PR updates the Gemfile to use `.ruby-version` as the single source of truth for the version of Ruby. Bundler was updated to support the [file option](https://github.com/rubygems/rubygems/pull/6876).


gileswells (approved):
LGTM

PR #331, opened 2024-01-31
[API-33844] - Add rake task for deeplink csv file

gileswells:

This will scan the events table and output all requests for sandbox access of ACG APIs into a csv with individualized deeplinks for consumers to retrieve the test user data.

### DIFF
--- a/app/models/test_user_email.rb
+++ b/app/models/test_user_email.rb
@@ -6,9 +6,9 @@ class TestUserEmail < ApplicationRecord
   def get_deeplinks
     user = User.where(email: email).first
     links = ''
-    links += single_link(user, 'community-care-eligibility', 'Community Care Eligibility API)', communityCare)
-    links += single_link(user, 'patient-health', 'Patient Health API (FHIR))', health)
-    links += single_link(user, 'benefits-claims', 'Benefits Claims API)', claims)
+    links += single_link(user, 'community-care-eligibility', 'Community Care Eligibility API', communityCare)
+    links += single_link(user, 'patient-health', 'Patient Health API (FHIR)', health)
+    links += single_link(user, 'benefits-claims', 'Benefits Claims API', claims)
     links += single_link(user, 'veteran-verification', 'Veteran Verification API', verification)
 
     links.delete_suffix('\\n')
--- a/app/models/test_user_email.rb
+++ b/app/models/test_user_email.rb
@@ -7,7 +7,6 @@ class TestUserEmail < ApplicationRecord
     user = User.where(email: email).first
     links = ''
     links += single_link(user, 'community-care-eligibility', 'Community Care Eligibility API)', communityCare)
-    links += single_link(user, 'clinical-health', 'Clinical Health API)', clinicalHealth)
     links += single_link(user, 'patient-health', 'Patient Health API (FHIR))', health)
     links += single_link(user, 'benefits-claims', 'Benefits Claims API)', claims)
     links += single_link(user, 'veteran-verification', 'Veteran Verification API', verification)
--- a/app/models/test_user_email.rb
+++ b/app/models/test_user_email.rb
@@ -4,14 +4,24 @@ class TestUserEmail < ApplicationRecord
   include ApplicationHelper
 
   def get_deeplinks
-    user = User.where(email: self.email).first
+    user = User.where(email: email).first
     links = ''
-    links += "[https://developer.va.gov#{generate_deeplink('veteran-verification', user)}](Veteran Verification API)\\n" if self.verification
-    links += "[https://developer.va.gov#{generate_deeplink('community-care-eligibility', user)}](Community Care Eligibility API)\\n" if self.communityCare
-    links += "[https://developer.va.gov#{generate_deeplink('clinical-health', user)}](Clinical Health API)\\n" if self.clinicalHealth
-    links += "[https://developer.va.gov#{generate_deeplink('patient-health', user)}](Patient Health API (FHIR))\\n" if self.health
-    links += "[https://developer.va.gov#{generate_deeplink('benefits-claims', user)}](Benefits Claims API)\\n" if self.claims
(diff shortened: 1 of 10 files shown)


lib/tasks/lpb.rake
```diff
@@ -523,4 +523,58 @@ namespace :lpb do
...
+      csv << %w[email links]
+      testUsers.each do |testUser|
+        csv << [testUser.email, testUser.get_deeplinks]
+      end
+    end
+    puts 'Created and populated test-users-with-deeplinks.csv file.'
+    csvFile = File.read(csvFileName)
+    s3 = AwsS3Service.new
+    bucket = ENV.fetch('TEST_USERS_BUCKET')
+    puts bucket
```
jw81, line 571: Is it safe to print this out?  Any thing that's logged out will get dumped into log gathering services.

gileswells: Good catch

db/schema.rb
```diff
@@ -272,6 +272,16 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "test_user_emails", force: :cascade do |t|
+    t.text "email", null: false
+    t.boolean "claims", default: false, null: false
+    t.boolean "clinicalHealth", default: false, null: false
```
jw81, line 278: Hmmmmm, is clinical health one of the apis we're supposed to account for?
From the [original ticket](https://jira.devops.va.gov/browse/API-32452?focusedId=2455079&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2455079) where we were asked to pull ACG emails, the only APIs they requested were ::
- Benefits Claims API
- Community Care Eligibility API
- Patient Health API (FHIR)
- Veteran Service History and Eligibility API

jw81: This [Slack discussion](https://lighthouseva.slack.com/archives/C063JSEKWSY/p1700074838159729?thread_ts=1700074397.114959&cid=C063JSEKWSY) seems to confirm that Clinical Health is not included in this effort.

app/models/test_user_email.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class TestUserEmail < ApplicationRecord
+  include ApplicationHelper
+
+  def get_deeplinks
+    user = User.where(email: self.email).first
+    links = ''
+    links += "[https://developer.va.gov#{generate_deeplink('veteran-verification', user)}](Veteran Verification API)\\n" if self.verification
```
jw81, line 9: Is this trying to output a markdown style link?
If so, I'm pretty sure it would need to be flipped.

So instead of ::
[http://www.blah.com](my cool link)

It would be ::
[my cool link](http://www.blah.com)
* haha, GH is rendering my markdown example.  Basically, you'd just have to swap the square brackets and parens

I could be misunderstanding what this is aiming to do though.

gileswells: You are absolutely correct there good sir. 🤦‍♂️

db/schema.rb
gileswells: The only reason that I had this here was because it is listed as ACG in the portal on prod. BUT, I shall remove because it makes absolutely 0 sense for it to be included.

app/models/test_user_email.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class TestUserEmail < ApplicationRecord
+  include ApplicationHelper
+
+  def get_deeplinks
+    user = User.where(email: email).first
+    links = ''
+    links += single_link(user, 'community-care-eligibility', 'Community Care Eligibility API)', communityCare)
+    links += single_link(user, 'patient-health', 'Patient Health API (FHIR))', health)
+    links += single_link(user, 'benefits-claims', 'Benefits Claims API)', claims)
```
jw81, line 11: Looks like the `name` argument on these 3 have trailing `)`'s.  Are those meant to be there?

lib/tasks/lpb.rake
```diff
@@ -523,4 +523,61 @@ namespace :lpb do
...
+                               }, unique_by: :email)
+        end
+      end
+    end
+    puts 'All required signups added to test_user_emails table.'
+    test_users = TestUserEmail.all
+    CSV.open(csv_file_name, 'w') do |csv|
+      csv << %w[email links]
+      test_users.each do |user|
+        csv << [user.email, testUser.get_deeplinks]
```
jw81, line 568: This should be `user` instead of `testUser`, right?

spec/support/vcr_cassettes/aws/s3_put_object.yml
```diff
@@ -0,0 +1,122 @@
+---
```
jw81, line 1: It looks like there's also a `GET` and a `POST` in this cassette.
Are those needed in this cassette?

lib/tasks/lpb.rake
```diff
@@ -523,4 +523,61 @@ namespace :lpb do
...
+        csv << [user.email, user.get_deeplinks]
+      end
+    end
+    puts 'Created and populated test-users-with-deeplinks.csv file.'
+    csv_file = File.read(csv_file_name)
+    s3 = AwsS3Service.new
+    bucket = ENV.fetch('TEST_USERS_BUCKET')
+    response = s3.put_object(bucket: bucket, key: csv_file_name, fileContents: csv_file, content_type: 'text/csv')
+    if response.etag
+      puts "File uploaded to s3://#{bucket}/#{csv_file_name}"
```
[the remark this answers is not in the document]
gileswells: I like it.

spec/support/vcr_cassettes/aws/s3_put_object.yml
gileswells: Cassettes can include multiple episodes and this one covers all the interactions that include even when local testing is required to grab the credentials from the `ecs-local-endpoints` docker continer.

jw81 (approved):
LGTM 👍

jw81 (approved):

PR #337, opened 2024-02-06
[API-33844] - Fix veteran verification url slug

gileswells:

Simple fix to incorrect url slug


Gemfile.lock
```diff
@@ -325,11 +325,11 @@ GEM
       net-protocol
```
jw81, line 325: If it were me, I'd leave the dependency update separate, just in case.  😉

gileswells: The security check blocks this PR from being able to merge. We could do the nokogiri PR merge first if we want to.

jw81: Ah, that's a great point about it blocking new merges.
I see you already handled that piece.  👍

lib/tasks/lpb.rake
```diff
@@ -567,13 +573,11 @@ namespace :lpb do
...
-          processed += 1
-        rescue
-          puts "Error caught with TestUserEmail#id:#{user.id}"
-          errors += 1
-        end
+        csv << [user.email, user.get_deeplinks]
+        processed += 1
+      rescue
+        puts "Error caught with TestUserEmail#id:#{user.id}"
+        errors += 1
```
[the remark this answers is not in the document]
jw81: Yep!  
If you use a `rescue` inside of a method or block, then the `begin....end` aren't necessary.

lib/tasks/lpb.rake
```diff
@@ -525,12 +525,18 @@ namespace :lpb do
...
-  task idmeEmailsExport: :environment do
+  task :idmeEmailsExport, [:before] => [:environment] do |_, args|
     require 'csv'
     require 'set'
+    if args.before.blank?
+      puts 'You must specify a before argument which is the first Event ID omitted from the export.'
+      next
+    end
+    puts 'Clear existing data from test_user_emails table'
+    ActiveRecord::Base.connection.execute('TRUNCATE test_user_emails RESTART IDENTITY')
```
jw81, line 536: Can you help me understand why we need to truncate the table?  🤔

gileswells: If we leave the existing table contents we would still have records that were previously not listed as unsubscribed in the output and then when we pull the report again in the future the unsubscribed boolean field would have no impact causing people to get an email they opted out of.

andresriveratoro (approved):

PR #340, opened 2024-02-20
[API-33897] - Delete utilities and promote to prod code

gileswells:
https://jira.devops.va.gov/browse/API-33897

Remove utilities routes and all legacy code related to the now abandoned promote to prod initiative that previously existed.


app/api/v0/consumers.rb
```diff
@@ -188,14 +186,14 @@ def subscribed?(consumers)
         requires :urlSlug, type: String
       end
       post 'test-user-data' do
-        header 'Access-Control-Allow-Origin', request.host_with_port
+        # header 'Access-Control-Allow-Origin', request.host_with_port
```
jw81, line 189: Probably want to bring this back for now, right?

app/api/v0/consumers.rb
```diff
@@ -188,14 +186,14 @@ def subscribed?(consumers)
...
         protect_from_forgery
 
         if validate_deeplink_hash(params[:userId], params[:hash])
           s3 = AwsS3Service.new
           bucket = ENV.fetch('TEST_USERS_BUCKET')
           key = ENV.fetch('TEST_USERS_OBJECT_KEY')
-          response = s3.get_object(bucket: bucket, key: key)
+          response = s3.get_object(bucket:, key:)
```
jw81, line 196: Was this a pattern you saw and really liked?  Or did Rubocop or something call it out?
I'm not opposed to the changes btw.  Just curious how ya came about it.    😄

lib/tasks/utilities.rake
```diff
@@ -0,0 +1,27 @@
...
+      api_environment = ApiEnvironment.find_by(environment:, api:)
+
+      users = api_environment.consumer_api_assignment.map do |record|
+        record.consumer.user.kept? ? record.consumer.user : nil
+      end.compact
+
+      puts users, with: Entities::UserEntity
+    else
+      users = User.kept.select { |user| user.consumer.present? }
+      puts users
```
jw81, line 24: I'm not sure we want to be printing out the list of `users` because anything we print is going to be revealed in logs.

jw81 (changes requested):
Threw a couple comments on there.  For one of them, I think we'll definitely have to remove some logic because it relies on printing to console and that data would likely be flagged as PII.

app/api/v0/consumers.rb
gileswells: Just Rubocop yelling at us in the builds. They were all automatically correctable so it still "passed" the builds but it was ugly.

app/api/v0/providers.rb
```diff
@@ -340,7 +340,7 @@ def create_internal_consumer(user)
               user = initialize_user
               kong_consumer = Kong::ServiceFactory.service(:sandbox).third_party_signup(user, api)
               present user, with: V0::Entities::ApikeyEntity,
-                            kong_consumer: kong_consumer,
```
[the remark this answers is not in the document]
gileswells: Checked and there isn't any usage on these endpoints. Deleted.

jw81 (approved):
LGTM 👍 

We'll wanna make sure to do some decent functional testing in the lower environments to make sure everything is still working as expected once this is merged in.

PR #347, opened 2024-03-20
[API-34321] - More UX updates to sandbox template

gileswells:
https://jira.devops.va.gov/browse/API-34321

More overall fixes to the email template including spoofing the ordered list with tables (thanks Outlook) and also removing some spacing. Deduplicated some text in the ACG portion of the template and added the last two lines that related to prod access and contacting support.


jw81 (approved):
LGTM 👍

PR #351, opened 2024-04-02
[API-35203] - Switch back to pulling Docker images from Dockerhub

jw81:
Original Issue :: [API-35203](https://jira.devops.va.gov/browse/API-35203)

- Switch back to pulling Docker images from Dockerhub instead of Amazon's ECR


gileswells (approved):
I love how silly a "fix" this is.

PR #355, opened 2024-04-10
[API-35427] - Add restricted_access field and data migration

gileswells:
https://jira.devops.va.gov/browse/API-35427

This PR updates the api_metadata table to include a JSON field for storing the different permutations of how different APIs can be listed as "Restricted Access."


db/migrate/20240405223356_api_restricted_access.rb
```diff
@@ -0,0 +1,71 @@
+class ApiRestrictedAccess < ActiveRecord::Migration[7.0]
```
jw81, line 1: Any reason why wouldn't want to handle this data transformation via a rails rake task?"
This just feels like a lot of logic with multiple points of possible failure to run via a DB migration.

db/migrate/20240405223356_api_restricted_access.rb
```diff
@@ -0,0 +1,71 @@
...
+  def up
+    require 'open-uri'
+    add_column :api_metadata, :restricted_access, :jsonb, default: "{\"restricted\":false}"
+
+    ApiMetadatum.find_each do |api|
+      if api['va_internal_only'] === 'StrictlyInternal'
```
jw81, line 7: Should really probably be accessing the attributes on these `ApiMetadatum` objects like this, `api.va_internal_only`.

If it's a column defined in the ActiveRecord/table schema, or a method defined in the model itself, then you should be able to access it via method calls, rather than using hash-like access.

db/migrate/20240405223356_api_restricted_access.rb
```diff
@@ -0,0 +1,71 @@
...
+        # For all StrictlyInternal APIs, block sandbox form and list as restricted
+        # This matches all current existing patterns
+        api['restricted_access'] = {
+          blockSandboxAccessForm: true,
+          restricted: true
+        }.to_json
+        api.save
+        next
+      end
+      if api['va_internal_only'] === 'AdditionalDetails' || api['va_internal_only'] === 'FlagOnly'
```
jw81, line 17: Ruby doesn't typically use the `===` for string comparison.
Rather, it's really only used for class matching, or range inclusion, or regex matching.

The ruby convention for this would be to just use `==`.

jw81 (dismissed):
Couple things to look over and consider.

db/migrate/20240405223356_api_restricted_access.rb
gileswells: I keep forgetting that I actually have console access through that large number of hops and jumpboxes to be able to run said rake task. I'll move it over to a rake task.

db/migrate/20240405223356_api_restricted_access.rb
gileswells: Wait, Rails does WHAT??? ❤️❤️❤️

andresriveratoro (approved):

PR #360, opened 2024-05-01
[API-35047] - Integrate with lighthouse auth for app creation

jw81:
Original Issue :: [API-35047](https://jira.devops.va.gov/browse/API-35047)

- Adds ability for LPB to make OAuth app creation requests via the Lighthouse Auth Service.
- This will allow LPB to stop making direct calls in OKTA


.ruby-version
```diff
@@ -1 +1 @@
-3.2.2
+3.2.3
```
jw81, line 1: Team BananaPeals [updated the version of ruby baked into the Base Ruby image](https://github.com/department-of-veterans-affairs/health-apis-docker-octopus/pull/201) we utilize.
This requires us to update the ruby version that LPB is targeting.

Gemfile
```diff
@@ -22,7 +22,7 @@ gem 'grape-swagger'
 gem 'grape-swagger-rails'
 gem 'httparty'
 gem 'jbuilder', '~> 2.7'
-gem 'jwt'
+gem 'jwt', '2.5.0'
```
jw81, line 25: The update to the base ruby image mentioned [here](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/360/files#r1586863879) lead to this.  The version of `jwt` gem we were using wasn't compatible with the version of ruby we were forced to move to.

app/api/validators/provided_oauth_params.rb
```diff
@@ -38,7 +38,7 @@ def validate_ccg_oauth_params!(params)
 
       begin
         JWT::JWK.import(jwk)
-      rescue ArgumentError
+      rescue JWT::JWKError
```
jw81, line 41: The update to the base ruby image mentioned [here](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/360/files#r1586863879) lead to this. We had to bump up to a newer version of the `jwt` gem and that new gem version changed what exception is raised.

app/api/v0/consumers.rb
```diff
@@ -150,7 +150,15 @@ def subscribed?(consumers)
...
-        okta_consumers = Okta::ServiceFactory.service(:sandbox).consumer_signup(user, okta_signup_options)
+
+        okta_consumers = if Flipper.enabled? :use_lighthouse_auth_service
+                           api_oauth_type = user.consumer.apis_list.first.auth_type
+                           oauth_app_requested = %w[acg ccg].include?(api_oauth_type)
+                           LighthouseAuth::AppCreate::Service.new.consumer_signup(user:, type: api_oauth_type, options: okta_signup_options) if oauth_app_requested # rubocop:disable Layout/LineLength
+                         else
+                           Okta::ServiceFactory.service(:sandbox).consumer_signup(user, okta_signup_options)
+                         end
+
```
jw81, line 161: Once we fully switch over to utilizing Lighthouse Auth Service, this `if..else..end` logic and all `Okta::` scoped logic will be removed.

app/services/lighthouse_auth/api_client.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module LighthouseAuth
+  class ApiClient
+    include HTTParty
+    base_uri Figaro.env.lighthouse_auth_service
+
```
jw81, line 7: I will add the following env vars to LPB's deployment unit repo and the AWS ParamStore before this is merged.
```
lighthouse_auth_service
lighthouse_auth_api_key
lighthouse_auth_access_key
```

I will follow up here with an update once I have completed that work.

build.sh
```diff
@@ -15,7 +15,7 @@ docker-compose -f docker-compose.yml -f docker-compose.dependencies.yml run app
 echo 'Building production container...'
 docker build --pull --file $BASEDIR/Dockerfile --target prod --tag $REPOSITORY:$VERSION $BASEDIR
 
-if [ $RELEASE == true ]; then
+if [ "${RELEASE:-false}" == "true" ]; then
```
jw81, line 18: This update was recommended by Team Shanktopus.

app/services/lighthouse_auth/app_create/service.rb
```diff
@@ -0,0 +1,50 @@
...
+          raise LighthouseAuthServiceErrors::InternalServerError, 'Internal server error'
+        else
+          raise LighthouseAuthServiceErrors::InternalServerError, "Unexpected status code #{response.code}"
+        end
+      end
+
+      def save_id_to_user(user:, type:, id:)
+        auth_ref_key = ConsumerAuthRef::KEYS["sandbox_#{type}_oauth_ref".to_sym]
+        existing_auth_ref = user.consumer.consumer_auth_refs.kept.find_by(key: auth_ref_key)
+        existing_auth_ref.discard if existing_auth_ref.present?
```
gileswells, line 42: I know this is how we're currently doing this but I'm not a fan of this. We're discarding the references and thus making the okta apps orphaned without good cause.

app/services/lighthouse_auth/app_create/service.rb
```diff
@@ -0,0 +1,50 @@
...
+          raise LighthouseAuthServiceErrors::NotFound, "Resource not found: #{response.body}"
+        when 500
+          raise LighthouseAuthServiceErrors::InternalServerError, 'Internal server error'
+        else
+          raise LighthouseAuthServiceErrors::InternalServerError, "Unexpected status code #{response.code}"
+        end
+      end
+
+      def save_id_to_user(user:, type:, id:)
+        auth_ref_key = ConsumerAuthRef::KEYS["sandbox_#{type}_oauth_ref".to_sym]
```
gileswells, line 40: Obviously right now we are just using this for sandbox only but having the ability to set the environment on this method would be good for the future. I'm ok leaving it as is for now.

lib/tasks/add_lh_auth_categories.rake
```diff
@@ -0,0 +1,87 @@
...
+  auscxrtjx2SgddXjG2p7: 'insurance/system',
+  ausdfm4dz4WYQ1UIT2p7: 'LCS',
+  auseavl6o5AjGZr2n2p7: 'loan-guaranty/system',
+  aushu53xo9vsmmmKv2p7: 'loan-review/system',
+  aus8x27nv4g4BS01v2p7: 'pgd/system',
+  ausftw7zk6eHr7gMN2p7: 'va-letter-generator/system',
+  ausfhzmx8hzZ6Pdye2p7: 'va-profile/system',
+  aus7y0sefudDrg2HI2p7: 'veteran-verification',
+  ausi3u00gw66b9Ojk2p7: 'veteran-verification/system'
+}.freeze
```
gileswells, line 28: Is there a reason for adding a new mapping in the oauth_info field using these values as opposed to pulling from the `sandboxAud` and `productionAud` values? My spot check shows that they line up with the already existing `sandboxAud` property. 

Also, how will this work if we eventually start managing prod access credentials through LPB when we get user accounts spun up?

lib/tasks/add_lh_auth_categories.rake
```diff
@@ -0,0 +1,87 @@
...
+    prod_category || lower_category
+  end
+  # rubocop:enable Metrics/MethodLength
+
+  def set_lighthouse_auth_service_category(oauth_type:, oauth_info:, category:)
+    case oauth_type
+    when :acg
+      oauth_info['acgInfo']['lighthouse_auth_service_category'] = category
+    when :ccg
+      oauth_info['ccgInfo']['lighthouse_auth_service_category'] = category
```
gileswells, line 83: If we still want to list off the lighthouse auth service category as it's own property in the JSON field then I would suggest going camel case so we match the casing of the existing JSON properties.
```suggestion
      oauth_info['acgInfo']['lighthouseAuthServiceCategory'] = category
    when :ccg
      oauth_info['ccgInfo']['lighthouseAuthServiceCategory'] = category
```

gileswells (changes requested):
Have some feedback but overall this looks very good.

app/services/lighthouse_auth/app_create/service.rb
jw81: Yeah, I was reluctant to stray away from what this was currently doing.
I guess, would you rather just leave the pre-existing record alone?
FWIW, that `.discard` is a soft-delete.  So it doesn't actually delete the record.  It just marks it as "discarded".

app/services/lighthouse_auth/app_create/service.rb
jw81: If you're good, I'd rather leave it as-is for now.
I just prefer to avoid adding "potential future" things until it's certain it'll be used/needed.

lib/tasks/add_lh_auth_categories.rake
jw81: From my discussions with @crolarlibertyva during development on this, I learned that moving forward, Team Pivot will no longer be providing us with the `sandboxAud`/`productionAud` values.  Instead, they'll be providing us the category + auth type values. (ex. `claims` and `claims/system`).  Correct me if I misunderstood anything @crolarlibertyva .

So this rake task should really only need to be ran once to set the new lighthouse auth service category for existing APIs today.  Then going forward, when we add new APIs, we would be making sure to set the lighthouse auth service category attribute instead of the `sandboxAud` and `productionAud` attributes.

lib/tasks/add_lh_auth_categories.rake
jw81: D'oh 🤦 
My ruby-mind took over there.  😜 
Yeah, that should follow the JS naming convention.  There's a couple places I'll need to update that at.  So I'll go add a commit to change all of them.

lib/tasks/add_lh_auth_categories.rake
crolarlibertyva: Correct.

lib/tasks/add_lh_auth_categories.rake
jw81: Made that change [here](https://github.com/department-of-veterans-affairs/lighthouse-platform-backend/pull/360/commits/2b01d7ba7565ccea160f1608e50e3f78c3dfa499). 

(of course the build is failing now due to random Docker rate-limiting 😑)

app/services/lighthouse_auth/app_create/service.rb
gileswells: Unfortunately, leaving the pre-existing record alone causes an issue where you cannot update a consumer record through rails_admin (short term problem but annoying all the same). Let's leave it and discuss how we want to address this after we get into the process of user accounts where they might genuinely have multiple client IDs of which they would need access to retrieve/update the credentials of.

gileswells (approved):

PR #361, opened 2024-05-06
API-35699 - Add "Organization" to sandbox signup slack message

jw81:
Original Issue :: [API-35699](https://jira.devops.va.gov/browse/API-35699)
- Adds the "Organization" value to the sandbox signup Slack alert


.ruby-version
```diff
@@ -1 +1 @@
-3.2.2
+3.2.3
```
jw81, line 1: Team BananaPeals [updated the version of ruby baked into the Base Ruby image](https://github.com/department-of-veterans-affairs/health-apis-docker-octopus/pull/201) we utilize.
This requires us to update the ruby version that LPB is targeting.

gileswells (approved):
LGTM

PR #366, opened 2024-05-10
[API-35913] - Sandbox signup reports endpoint

gileswells:
https://jira.devops.va.gov/browse/API-35913

This endpoint will be used to expose sandbox signup events for the Salesforce surge team to consume for their data syncing purposes. This is the proposed output which the Salesforce team has already agreed to being able to consume. Changes can be made but they would need good reasoning as to why it _needs_ to change.

```
{
  "signups": [
    {
      "signup": {
        "signupId": 119,
        "createdAt": "2024-05-06T17:36:32.533Z",
        "user": {
          "id": 1,
          "firstName": "Giles",
          "lastName": "Wells",
          "email": "giles.wells@oddball.io",
          "emailUnsubscribed": false
        },
        "organization": "Oddball",
        "apis": [
          {
            "api": {
              "ref": "facilities",
              "name": "VA Facilities API",
              "authTypes": [
                {
                  "authType": {
                    "authType": "apikey",
                    "consumerId": "0aba62b4-ca5b-4fda-997e-67012be40196"
                  }
                }
              ]
            }
          },
          {
            "api": {
              "ref": "health",
              "name": "Patient Health API (FHIR)",
              "authTypes": [
                {
                  "authType": {
                    "authType": "ccg",
                    "clientId": "0oas6sdg1tnAH3R3A2p7"
                  }
                },
                {
                  "authType": {
                    "authType": "acg",
                    "clientId": "0oaray2sifTb7Jur42p7",
                    "deeplinkUrl": ""
                  }
                }
              ]
            }
          }
        ]
      }
    }
  ]
}
```


app/api/v0/consumers.rb
```diff
@@ -95,6 +95,25 @@ def subscribed?(consumers)
         present consumers, with: V0::Entities::ConsumerEntity
       end
 
+      desc 'Report of consumer signups'
+      params do
+        requires :after, type: Integer, allow_blank: false
+        requires :before, type: Integer, allow_blank: false
```
jw81, line 101: Is this milliseconds or seconds since epoch?

gileswells: Just seconds. Since it's going to be based on an hourly pull I don't think there's any reason to do milliseconds.

app/api/v0/consumers.rb
```diff
@@ -95,6 +95,25 @@ def subscribed?(consumers)
...
+        validate_token(Scope.consumer_read)
+        require 'date'
+
+        before_time = Time.zone.at(params[:before]).to_datetime
+        after_time = Time.zone.at(params[:after]).to_datetime
+
+        events = Event
+                 .where(event_type: Event::EVENT_TYPES[:sandbox_signup])
+                 .where(['created_at <= :before', { before: before_time }])
+                 .where(['created_at > :after', { after: after_time }])
```
jw81, line 113: Not saying it needs to change, just wanted to mention that this should work too and is a little shorter.

```
events = Event
          .where(event_type: Event::EVENT_TYPES[:sandbox_signup])
          .where('created_at <= ?', before_time)
          .where('created_at > ?', after_time)
```

app/api/v0/consumers.rb
```diff
@@ -95,6 +95,25 @@ def subscribed?(consumers)
...
       end
 
+      desc 'Report of consumer signups'
+      params do
+        requires :after, type: Integer, allow_blank: false
+        requires :before, type: Integer, allow_blank: false
+      end
+      get '/signups' do
+        validate_token(Scope.consumer_read)
+        require 'date'
```
jw81, line 105: I'm surprised that Rails isn't auto requiring this.  🤔 
If you remove this, is Rails unable to use the library?

app/api/v0/consumers.rb
gileswells: I like yours better

app/api/v0/consumers.rb
gileswells: Yeah... it works without. I was having issues with date matching in the spec tests.

app/api/v0/entities/consumer_signup_entity.rb
```diff
@@ -0,0 +1,24 @@
...
+  module Entities
+    class ConsumerSignupEntity < Grape::Entity
+      expose :id, as: :signupId
+      expose :created_at, as: :createdAt
+      expose :user do |event|
+        user = User.find_by!({ email: event.content['email'] })
+        V0::Entities::UserSignupEntity.represent(user)
+      rescue
+        { error: 'User Not Found' }
+      end
```
jw81, line 13: I wonder if we should explicitly handle the case where a user can't be found and then have a more generic way to handle any other possible error.

Something like,
```
expose :user do |event|
  begin
    user = User.find_by!(email: event.content['email'])
    V0::Entities::UserSignupEntity.represent(user)
  rescue ActiveRecord::RecordNotFound
    { error: 'User Not Found' }
  rescue StandardError => e
    { error: 'An error occurred', details: e.message }
  end
end
```

What do you think?

gileswells: My only issue with adding this in is that I'd have to do some crazy test to get this additional rescue to fire. I realize that I've missed covering the original rescue in my tests already (fixing that now) but what am I going to do to get that next rescue to fire?

app/api/v0/consumers.rb
```diff
@@ -95,6 +95,23 @@ def subscribed?(consumers)
         present consumers, with: V0::Entities::ConsumerEntity
       end
 
+      desc 'Report of consumer signups'
+      params do
+        requires :after, type: Integer, allow_blank: false
+        requires :before, type: Integer, allow_blank: false
+      end
+      get '/signups' do
+        validate_token(Scope.consumer_read)
```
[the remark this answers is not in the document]
gileswells: Part of me wanted to do that in a new PR. I was having enough issues with getting the unit tests to pass consistently I kinda wanted to act like an ostrich and bury my head in the sand. I'll get this sorted.

app/api/v0/entities/consumer_signup_entity.rb
jw81: It should be possible to mock the behavior of one of the calls in the `begin` and force it to raise a `StandardError`.

So something like this could work.  This is psuedo-ish code.  So fair chance it would need some tweaks.
```
   context 'when an unexpected error is encountered' do
    before do
      allow(V0::Entities::UserSignupEntity).to receive(:represent).and_raise(StandardError, 'Something went wrong')
    end
	
    it 'returns an error message' do
      signup_params[:apis] = "acg/#{api_ref_two},ccg/#{api_ref_two}"
      VCR.use_cassette('okta/consumer_signup_200', match_requests_on: [:method]) do
        VCR.use_cassette('okta/consumer_signup_ccg_200', match_requests_on: [:method]) do
          post apply_base, params: signup_params
        end
      end
	  
      get '/platform-backend/v0/consumers/signups?after=1715016990&before=2715016994'

      expect(response.body).to eq({ error: 'An error occurred', details: 'Something went wrong' })
    end
   end
  ```

File: app/api/base.rb [hunk 6]
```diff
@@ -61,6 +61,17 @@ def protect_from_forgery
       end
     end
 
+    def validate_apikey
+      raise AuthorizationError unless Flipper.enabled? :allow_apikey_validation
+      return unless Flipper.enabled? :validate_apikey
```
jw81, line 66: I'm trying to understand why we need this additional flipper.  🤔

gileswells: So, without this flipper, the endpoint would be available on all environments. We could do something that's more strictly tied to the environment but that makes testing more ugly.
Essentially on dev we'd turn on allow_apikey_validation and validate_apikey. This also makes it more convenient to use it locally because just allow_apikey_validation means we can hit the endpoint without an apikey in the request header.

jw81: Oh, does it exist solely to prevent the need to duplicate the apikey header on each spec?

jw81: Just closing the loop here.
We chatted offline and landed on an approach that uses a single feature flag to enable/disable the entire endpoint.
If enabled, apikey validation will take place and the endpoint can continue.
If disabled, we'll bomb out of the endpoint immediately and not allow the endpoint to execute.

jw81 (approved):
LGTM 👍

PR #370, opened 2024-05-21
[API-36639] - Create task for servicing sitemap urls

gileswells:
https://jira.devops.va.gov/browse/API-36639

While not explicitly called out for in the ticket, this new rake task will allow us to seed and re-seed the static urls that are included in the [sitemap.xml](https://developer.va.gov/platform-backend/sitemap.xml) file / endpoint that we use for the public portal. This rake task will allow us to forego making an internal API endpoint to manage adding and removing static page routes from the sitemap.


jw81 (approved):
LGTM 👍

